PR #13, opened 2015-12-19
crypto/tls: implement countermeasures against CBC padding oracles

FiloSottile:


FiloSottile:
https://go-review.googlesource.com/18130

PR #14, opened 2016-07-17
Implementation of Chacha20 and Poly1305 for TLS

FiloSottile:
It uses the same high speed implementation, used by CF in OpenSSL


src/crypto/tls/cipher_suites.go
```diff
@@ -75,6 +76,8 @@ type cipherSuite struct {
 var cipherSuites = []*cipherSuite{
 	// Ciphersuite order is chosen so that ECDHE comes before plain RSA
 	// and RC4 comes before AES-CBC (because of the Lucky13 attack).
+	{TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305, 32, 0, 12, ecdheRSAKA, suiteECDHE | suiteTLS12, nil, nil, aeadChacha20Poly1305},
+	{TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305, 32, 0, 12, ecdheECDSAKA, suiteECDHE | suiteECDSA | suiteTLS12, nil, nil, aeadChacha20Poly1305},
```
FiloSottile: Not sure ChaCha20 should come before AES ciphersuites.

It would be nice to do the detect-client-preference thing, but let's leave that for later.

src/crypto/tls/cipher_suites.go
```diff
@@ -143,6 +146,11 @@ type fixedNonceAEAD struct {
 	aead                 cipher.AEAD
 }
 
+type xoredNonceAEAD struct {
+	fixedNonce []byte
+	aead       cipher.AEAD
+}
+
```
FiloSottile: This belongs below before its methods.

src/crypto/tls/cipher_suites.go
```diff
@@ -173,6 +206,18 @@ func aeadAESGCM(key, fixedNonce []byte) cipher.AEAD {
 	return &fixedNonceAEAD{nonce1, nonce2, aead}
 }
 
+func aeadChacha20Poly1305(key, fixedNonce []byte) cipher.AEAD {
```
FiloSottile: Wondering if this should be [Cc]haCha.

src/crypto/tls/handshake_client_test.go
```diff
@@ -296,6 +296,7 @@ func (test *clientTest) run(t *testing.T, write bool) {
 	if config == nil {
 		config = testConfig
 	}
+	config.CipherSuites = allCipherSuites()[2:]
```
FiloSottile: Instead of doing this, we can regenerate the test vectors.

And we should add tests for ChaCha.

src/crypto/chacha20poly1305/chacha20.go
```diff
@@ -0,0 +1,169 @@
...
+)
+
+type chacha20Cipher struct {
+	state [16]uint32
+	buf   [16]uint32
+	avail int
+}
+
+type KeySizeError int
+type IVSizeError int
```
FiloSottile: Types definitions go together with their methods.

src/crypto/chacha20poly1305/poly1305.go
```diff
@@ -0,0 +1,181 @@
...
+
+func writeU32(in uint64, out []byte) {
+	/* Assumption : len(out) >= 8 */
+	for i := 0; i < 4; i++ {
+		out[i] = byte(in)
+		in >>= 8
+	}
+}
+
+func NewMac(key []byte) (*poly1305MAC, error) {
```
FiloSottile: Don't return private structs from public functions.

src/crypto/chacha20poly1305/xor.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2013 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
FiloSottile: I guess this file is duplicated somewhere, we should considering moving it to an internal package for sharing.

FiloSottile:
Splitting this into one commit for the chacha20poly1305 package and one for the TLS addition.

FiloSottile:
(Sorry for the PEDANTIC=1 review, I assume we are upstreaming this, so I'm applying stdlib standards. I'll take care of some of the comments.)

src/crypto/tls/cipher_suites.go
FiloSottile: Yep.

> RFC 7539 - ChaCha20 and Poly1305 for IETF Protocols

src/crypto/chacha20poly1305/chacha20.go
```diff
@@ -0,0 +1,169 @@
...
+	c.state[12] = ctr
+	c.state[13] = u8tou32(iv[0:4])
+	c.state[14] = u8tou32(iv[4:8])
+	c.state[15] = u8tou32(iv[8:12])
+	c.avail = 0
+
+	return nil
+}
+
+func NewCipher(key, iv []byte, ctr uint32) (cipher.Stream, error) {
```
FiloSottile: This should return a Cipher object instead of a interface, so that users have a concrete type to type asset to, for example.

vkrasnov: That what cipher.NewCTR returns

FiloSottile: Hmm, true, but all the cipher-specific ones have a concrete (look at rc4). I think the reason cipher.NewCTR return a interface is that it's a wrapper for an interface itself (cipher.Block).

src/crypto/tls/cipher_suites.go
FiloSottile: @vkrasnov About having ChaCha first, how do the benchmarks stack up?

FiloSottile:
@vkrasnov I added my changes in #15 (because I haven't gotten to the TLS part yet)

src/crypto/tls/cipher_suites.go
```diff
@@ -156,6 +164,31 @@ func (f *fixedNonceAEAD) Open(out, nonce, plaintext, additionalData []byte) ([]b
 	return f.aead.Open(out, f.openNonce, plaintext, additionalData)
 }
 
+func (f *xoredNonceAEAD) NonceSize() int { return 12 }
```
FiloSottile: > // NonceSize returns the size of the nonce that must be passed to Seal
> // and Open.

So 8, right?

vkrasnov: Yes in that case it is 8.  8 are passed to xoredNonceAEAD that are then made into 12 for ChaCha-Poly.

PR #16, opened 2016-08-16
Implement basic server push

vkrasnov:


FiloSottile:
The http2 code is bundled from golang.org/x/net/http2 so it should be modified there.

```
 // Code generated by golang.org/x/tools/cmd/bundle.
 //go:generate bundle -o h2_bundle.go -prefix http2 golang.org/x/net/http2
```

src/go/build/deps_test.go
```diff
@@ -382,6 +382,7 @@ var pkgDeps = map[string][]string{
 		"golang.org/x/net/lex/httplex",
 		"internal/nettrace",
 		"net/http/httptrace",
+		"regexp",
```
FiloSottile: This will not fly upstream. Can we do without?

vkrasnov: There is not enough consensus on server push and how to do it, for this code to be even considered for upstream.
For now people can use it from our repo.

vkrasnov: > The http2 code is bundled from golang.org/x/net/http2 so it should be modified there.

I don't see any real reason to do it that way, since like I said, its not going upstream.

thibmeu:
Golang has introduced [Pusher](https://golang.org/pkg/net/http/#Pusher) interface. It can be used to implement server push.
In addition, there are [discussions](https://groups.google.com/a/chromium.org/g/blink-dev/c/K3rYLvmQUBY/m/vOWBKZGoAQAJ) about removing it from Chromium.
Developers should look at using [103 Early Hints status code](https://tools.ietf.org/html/rfc8297).

henrydcase:
Good!

PR #19, opened 2016-09-09
Integrate the x25519 implementation from x/crypto

vkrasnov:
Enable it in TLS as the preferred ECDHE curve, as it has comparable performance to P-256, but is inherently safer


ogud:
Are we sure?
I have doubt's about the sanity of the proponent of this algorithm

Ólafur

sent from phone

FiloSottile:
Me, Vlad or DJB? Because could be any of the three. 😜

ogud:
DJB is the one I do not trust

Olafur

PR #20, opened 2016-09-28
Check input length when unmarshaling non-WS curves

ghedo:


src/crypto/elliptic/elliptic.go
```diff
@@ -339,7 +339,10 @@ func Unmarshal(curve Curve, data []byte) (x, y *big.Int) {
 	byteLen := (curve.Params().BitSize + 7) >> 3
 
 	if _, ok := curve.(nonWSCurve); ok {
-		x = new(big.Int).SetBytes(data[0:byteLen])
+		if len(data) != byteLen {
+			return
+		}
+		x = new(big.Int).SetBytes(data[:byteLen])
```
FiloSottile: No need for `[:byteLen]` here.

vkrasnov: I think so too

ghedo: Ok, fixed (I think that's what you meant anyway).

FiloSottile:
Thanks!

PR #21, opened 2016-11-17
Catch when RemoteAddr() returns nil.

Bren2010:

### DIFF
--- a/src/net/http/server.go
+++ b/src/net/http/server.go
@@ -1483,7 +1483,13 @@ func (e badRequestError) Error() string { return "Bad Request: " + string(e) }
 
 // Serve a new connection.
 func (c *conn) serve(ctx context.Context) {
-	c.remoteAddr = c.rwc.RemoteAddr().String()
+	raddr := c.rwc.RemoteAddr()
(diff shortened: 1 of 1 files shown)


src/net/http/server.go
```diff
@@ -1483,7 +1483,13 @@ func (e badRequestError) Error() string { return "Bad Request: " + string(e) }
 
 // Serve a new connection.
 func (c *conn) serve(ctx context.Context) {
-	c.remoteAddr = c.rwc.RemoteAddr().String()
+	raddr := c.rwc.RemoteAddr()
+	if raddr == nil {
+		c.setState(c.rwc, StateClosed)
```
FiloSottile: No `c.close()`?

FiloSottile:
Is there a deeper root cause? Why does this happen? I'm sure this will mute it, but why the hell is it reached? Also, please open a Go issue.

Lekensteyn:
My guess is that this is a custom `net.Conn` implementation that somehow returned `nil`. Shall we close this PR since that can usually not happen for standard implementations?

PR #24, opened 2020-07-08
tls: add support for Ed25519-Dilithium3

bwesterb:
Notes

- The `cf-copy-from-circle.py` copies circl (including Dilithium) into stdlib. As Circl's ed25519 isn't yet a drop-in replacement for Go's ed25519, it's more convenient to copy all of Circl into the stdlib than to rewrite the API on the fly.
- The stdlib tls tests are generated by openssl.  As openssl does not (yet) support Ed25519-Dilithium3, we can't regenerate the test data easily.
- [Example of usage](https://gist.github.com/bwesterb/621ac21e5f65ff9587ad6b00a2c41a99)


armfazh (approved):

src/crypto/tls/generate_cert.go
```diff
@@ -28,13 +29,14 @@ import (
...
-	ecdsaCurve = flag.String("ecdsa-curve", "", "ECDSA curve to use to generate a key. Valid values are P224, P256 (recommended), P384, P521")
-	ed25519Key = flag.Bool("ed25519", false, "Generate an Ed25519 key")
+	host            = flag.String("host", "", "Comma-separated hostnames and IPs to generate a certificate for")
+	validFrom       = flag.String("start-date", "", "Creation date formatted as Jan 1 15:04:05 2011")
+	validFor        = flag.Duration("duration", 365*24*time.Hour, "Duration that certificate is valid for")
+	isCA            = flag.Bool("ca", false, "whether this cert should be its own Certificate Authority")
+	rsaBits         = flag.Int("rsa-bits", 2048, "Size of RSA key to generate. Ignored if --ecdsa-curve is set")
+	ecdsaCurve      = flag.String("ecdsa-curve", "", "ECDSA curve to use to generate a key. Valid values are P224, P256 (recommended), P384, P521")
+	ed25519Key      = flag.Bool("ed25519", false, "Generate an Ed25519 key")
+	eddilithium3Key = flag.Bool("eddilithium3", false, "Generate an Ed25519-Dilithium3 key")
```
claucece, line 39: While this is nice, it is difficult to see that it is 'ed25519 + dilithium', and might got confused if added with ed448...

src/crypto/x509/pkcs8.go
```diff
@@ -74,6 +76,20 @@ func ParsePKCS8PrivateKey(der []byte) (key interface{}, err error) {
 		}
 		return ed25519.NewKeyFromSeed(curvePrivateKey), nil
 
+	case privKey.Algo.Algorithm.Equal(oidPublicKeyEdDilithium3):
+		if l := len(privKey.Algo.Parameters.FullBytes); l != 0 {
+			return nil, errors.New("x509: invalid Ed25519-Dilithium3 private key parameters")
+		}
+		var packedSk []byte
```
claucece, line 83: I think they name in Golang secret keys 'Private Keys', so maybe is better that name.

src/crypto/x509/pkcs8.go
```diff
@@ -74,6 +76,20 @@ func ParsePKCS8PrivateKey(der []byte) (key interface{}, err error) {
...
+		if _, err := asn1.Unmarshal(privKey.PrivateKey, &packedSk); err != nil {
+			return nil, fmt.Errorf("x509: invalid Ed25519-Dilithium3 private key: %v", err)
+		}
+		var sk eddilithium3.PrivateKey
```
claucece, line 87: Same for this.

claucece:
Mmm... @bwesterb , I was  wondering if we could make the python script more multipurpose. And I was wondering also, on the other hand, if we should copy circl to 'src/circl'.. maybe we should copy to 'src/crypto/circl'? What do you think?

bwesterb:
> Mmm... @bwesterb , I was wondering if we could make the python script more multipurpose. 

What extra purposes were you thinking about?

> And I was wondering also, on the other hand, if we should copy circl to 'src/circl'.. maybe we should copy to 'src/crypto/circl'? What do you think?

I would really like to have dilithium under `crypto/dilithium`.  This was the approach I took earlier, but unfortunately, because `ed25519` isn't a drop-in replacement yet, I have to use the API of circl, which cascades: I then also have to pull in various internal circl packages such as `conv`.  So pulling in circl like this is a pragmatic temporary approach.

I think `crypto/circl/sign/eddilithium4 is a bit too long.

Also, I like just using `circl` because then it's clear it's something outside of the standard library.  Perhaps we should even use `cf/circl/...` or `cf/sign/...` to make that more clear.  @armfazh any opinions?

src/crypto/tls/generate_cert.go
bwesterb: I changed the flag.

src/crypto/x509/pkcs8.go
bwesterb: In APIs I use PrivateKey/PublicKey, but in function bodies having the `sk` and `pk` abbreviations is just so convenient as compared to `pubKey` and `privKey`.

claucece:
> What extra purposes were you thinking about?

Well, if I get correctly the python, it seems like the script is removing the circl dir and adding again. While this is nice, I was thinking that maybe the user of the script can pass what it wants to update as a argument to the python call, like `python cf-copy-from-circle.py update=ed448`, or something like that, so it doesn't have to be adding and readding each time.

> I think `crypto/circl/sign/eddilithium4 is a bit too long.

It is indeed long. But the reason why I like it is due to the fact that the 'crypto' stuff is on the 'crypto' folder, and might be easier to work with if we ever want to push this to upstream. But this can wait for when the API for ed25519 is ready to be a replacement.

src/crypto/x509/pkcs8.go
claucece: Yes, but as you can see on the ed25519 case, the golang people seem to be using (consistently everywhere else) the `privateKey` term as in `var curvePrivateKey []byte`. Might be good to keep that same style, and not diverge from the style of upstram.

bwesterb:
> > What extra purposes were you thinking about?
> 
> Well, if I get correctly the python, it seems like the script is removing the circl dir and adding again. While this is nice, I was thinking that maybe the user of the script can pass what it wants to update as a argument to the python call, like `python cf-copy-from-circle.py update=ed448`, or something like that, so it doesn't have to be adding and readding each time.

I think the majority of the time is spent in cloning the repository — the copying afterwards should be relatively fast.  We could cache the checkout of the Circl repository.  But, please, make any changes you like :).

> > I think `crypto/circl/sign/eddilithium4 is a bit too long.
> 
> It is indeed long. But the reason why I like it is due to the fact that the 'crypto' stuff is on the 'crypto' folder, and might be easier to work with if we ever want to push this to upstream. But this can wait for when the API for ed25519 is ready to be a replacement.

Yes, I would also like it in a form where we could upstream it.  However, with the "circle" in these we'll never get it upstream, so we'll have to wait for compatibility :).

armfazh:
Wouldn't be better to use circl as a submodule? This eases fetching the sources from the repo.
On the other hand, I think the path "src/crypto/circl" is appropriate for now.

bwesterb:
> Wouldn't be better to use circl as a submodule? This eases fetching the sources from the repo.
> On the other hand, I think the path "src/crypto/circl" is appropriate for now.

That would work, but we'll still need a script to copy and modify the import paths.  (Go stdlib can't use external modules.)

claucece:
> But, please, make any changes you like :).

Sure thing ;). Thanks!

> On the other hand, I think the path "src/crypto/circl" is appropriate for now.

Nice! Should we change this to "src/crypto/circl" then?

src/crypto/tls/handshake_server_test.go
```diff
@@ -684,9 +684,11 @@ func (test *serverTest) run(t *testing.T, write bool) {
...
 				t.Fatalf("%s #%d: %s\nRead %d, wanted %d, got %x, wanted %x\n", test.name, i+1, err, n, len(bb), bb[:n], b)
 			}
-			if !bytes.Equal(b, bb) {
-				t.Fatalf("%s #%d: mismatch on read: got:%x want:%x", test.name, i+1, bb, b)
-			}
+			// XXX disabling for now as we can't generate testvectors from openssl
+			//     as it doens't support EdDilithium3 yet.
+			// if !bytes.Equal(b, bb) {
+			// 	t.Fatalf("%s #%d: mismatch on read: got:%x want:%x", test.name, i+1, bb, b)
+			// }
```
thomwiggers, line 691: Doesn't this disable most of the client test suite?

bwesterb: Yes, it does.  I fixed that in a later version, but pushed it to the wrong repository.  It's updated now.

src/circl/.etc/docker/debian-buster-aarch64/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM multiarch/debian-debootstrap:arm64-buster
```
thomwiggers, line 1: Maybe files in the `.etc` folder shouldn't be copied by the script.

src/circl/Makefile
```diff
@@ -0,0 +1,51 @@
+# I'm sure there is better way. But I would need to find it first
```
thomwiggers, line 1: Is this file still necessary?

src/circl/codecov.yml
```diff
@@ -0,0 +1,21 @@
+# Default codecoverage configuration https://docs.codecov.io/docs/codecov-yaml
```
thomwiggers, line 1: This one can probably also be deleted.

bwesterb: Removed all of them.

bwesterb:
Superseded by #25

PR #29, opened 2020-08-23
crypto/tls/ech: ECH minting

cjpatton:
This changes adds a package called crypto/tls/ech that's used to generate ECH config/key pairs. It also implements the operations that are needed in the TLS stack.


cjpatton:
I'm dropping this PR in favor of #30: the crypto/tls/ech has been moved to github.com/cjpatton/ech-mint.

PR #34, opened 2020-10-16
Update circl

bwesterb:


armfazh:
Im ok with this, but why not to use go modules for tracking this dependency?

armfazh (approved):

bwesterb:
> Im ok with this, but why not to use go modules for tracking this dependency?

I don't understand what you propose.

PR #35, opened 2020-11-04
crypto/tls: Improve ECH metrics collection

cjpatton:
* Removes `ECHStatus` and `ECHOffered` from `ConnectionState`.

* Adds a callback `EventHandler()` to `Config`, which can be called at
various points during the handshake to respond to various events. For
example, this callback can be used to record metrics.

* Adds calls to `EventHandler()` just before closing the TLS connection
for resolving ECH usage: whether the client offered, greased, or
bypassed ECH; and whether the server accepted, rejected, or bypassed
ECH.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -315,27 +315,9 @@ type ConnectionState struct {
 	// RFC 7627, and https://mitls.org/pages/attacks/3SHAKE#channelbindings.
 	TLSUnique []byte
 
-	// ECHStatus reports the status of this connection's usage of the ECH
-	// extension.
-	ECHStatus ECHStatus
-
-	// ECHAOffered indicates whether the client offered a real (i.e., non-dummy)
-	// ECH extension in this connection.
-	ECHOffered bool
-
 	// ekm is a closure exposed via ExportKeyingMaterial.
 	ekm func(label string, context []byte, length int) ([]byte, error)
 }
-
-// ECHStatus represents the connection's ECH usage status.
-type ECHStatus uint8
-
-const (
-	ECHStatusBypassed ECHStatus = 0
-	ECHStatusAccepted           = 1
-	ECHStatusRejected           = 2
-	ECHStatusGrease             = 3
-)
 
 // ExportKeyingMaterial returns length bytes of exported key material in a new
 // slice as defined in RFC 5705. If context is nil, it is not used as part of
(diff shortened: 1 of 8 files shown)


src/crypto/tls/common.go
```diff
@@ -747,6 +729,11 @@ type Config struct {
 	// payload.
 	ServerECHProvider ECHProvider
 
+	// EventHandler, if set, is called by the client and server at various
+	// points during the handshake to handle specific events. This callback can
+	// be used to record interesting metrics.
+	EventHandler func(event string)
```
cjpatton, line 735: I'd imagine we want something more general than a `string` for the event's type. This might be better:
```
type Event interface {
    String() string
}
```

chris-wood: Yeah, or perhaps even: 

```go
type Event interface {
   Type() int
   Name() string
}

type CounterEvent struct {
   Count int
}

... to process an event ... 
if event.Type() == CounterEvent {
   counterEvent, ok := event.(CounterEvent)
   if !ok { panic(...) }
   // use counterEvent as needed
}
```

Sort of like [BoringSSL's info callback mechanism](https://github.com/google/boringssl/blob/master/include/openssl/ssl.h#L3976)

wbl: You don't need .Type: it's possible to introspect based on types in go.

chris-wood: Whatever's most idiomatic :) If you want to cast-then-check, that works too.

cjpatton: I went with
```
type Event interface {
   Name() string
}
```
I don't see a reason to do anything fancier. (We can always change this later.)

src/crypto/tls/common.go
```diff
@@ -747,6 +729,11 @@ type Config struct {
 	// payload.
 	ServerECHProvider ECHProvider
 
+	// EventHandler, if set, is called by the client and server at various
+	// points during the handshake to handle specific events. This callback can
+	// be used to record interesting metrics.
```
chris-wood, line 734: We should say something about the, uh, stability of these events? For example:

```suggestion
	// be used to record interesting metrics. Callers MUST NOT rely on the presence or order of 
	// events generated by this callback, as they may change at any time.
```

In a former life, clients ossified on such functionality and caused *all sorts of damage*.

chris-wood (approved):
This is really nice! Would you be open to extending this in the future to trace handshake stage events, e.g., receipt of CH, completion of the server's flight, etc.? (I could thread in that change.)

src/crypto/tls/common.go
cjpatton: - Added a note about this API being experimental.
- Prefixed `Event` and `Config.EventHandler` with `EXP_` in order to signal callers that they shouldn't rely on this API yet.

cjpatton:
> This is really nice! Would you be open to extending this in the future to trace handshake stage events, e.g., receipt of CH, completion of the server's flight, etc.? (I could thread in that change.)

That's a great idea. I think we need to emphasize that the API is unstable ... that should be clear now.

cjpatton:
Squashed and rebased.

PR #36, opened 2020-11-05
crypto/tls: Fix a bug in the ECH extension

cjpatton:
To compute ClientHelloOuterAAD, the server unmarshals ClientHelloOuter,
removes the ECH extension, then marshal the result the same way the
client does in this package. As a result, the server may re-order and
remove unrecognized extensions in the processed. This changes the
computation of ClientHelloOuterAAD to operate on the serialized
ClientHelloOuter.

### DIFF
--- a/src/crypto/tls/ech.go
+++ b/src/crypto/tls/ech.go
@@ -377,12 +377,15 @@ func (c *Conn) echAcceptOrBypass(hello *clientHelloMsg) (*clientHelloMsg, error)
 		return nil, fmt.Errorf("ech: %s", err)
 	}
 
-	var helloOuterAad clientHelloMsg
-	helloOuterAad = *hello
-	helloOuterAad.raw = nil
-	helloOuterAad.encryptedClientHello = nil
-	helloOuterAad.encryptedClientHelloOffered = false
-	rawEncodedHelloInner, err := ctx.open(helloOuterAad.marshal(), ech.payload)
+	helloOuterAad, ok := encodeClientHelloOuterAAD(hello.raw, extensionECH)
+	if !ok {
+		// This occurs if the ClientHelloOuter is malformed. This values was
+		// already parsed into `hello`, so this should not happen.
+		c.sendAlert(alertInternalError)
+		return nil, fmt.Errorf("ech: failed to compute ClientHelloOuterAAD")
+	}
+
+	rawEncodedHelloInner, err := ctx.open(helloOuterAad, ech.payload)
 	if err != nil {
 		if c.hrrTriggered && c.ech.accepted {
 			// Don't reject after accept, as this would result in processing the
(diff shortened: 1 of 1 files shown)


src/crypto/tls/ech.go
```diff
@@ -882,6 +885,85 @@ func echIsValidVersion(ext uint16) bool {
 	return ext == extensionECH
 }
 
+// encodeClientHelloOuterAAD interprets data as ClientHelloOuter and maps it to
+// ClientHelloOuterAAD. Returns a bit indicated whether parsing ClientHelloOuter
```
chris-wood, line 889:
```suggestion
// ClientHelloOuterAAD. Returns a bool indicated whether parsing ClientHelloOuter
```

src/crypto/tls/ech.go
```diff
@@ -882,6 +885,85 @@ func echIsValidVersion(ext uint16) bool {
...
+
+	if !s.ReadUint16(&legacyVersion) ||
+		!s.ReadBytes(&random, 32) ||
+		!s.ReadUint8LengthPrefixed(&legacySessionId) ||
+		!s.ReadUint16LengthPrefixed(&cipherSuites) ||
+		!s.ReadUint8LengthPrefixed(&legacyCompressionMethods) {
+		return nil, false
+	}
+
+	if !s.Empty() &&
```
chris-wood, line 919: Shouldn't extensions be non-empty? (Meaning, should fail if `s.Empty() == true` here?)

src/crypto/tls/ech.go
```diff
@@ -882,6 +885,85 @@ func echIsValidVersion(ext uint16) bool {
...
+		(!s.ReadUint16LengthPrefixed(&extensions) || !s.Empty()) {
+		return nil, false
+	}
```
chris-wood, line 922: Should we assert that `s.Empty() == true` at this point? (We've reached the end of the CH.)

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -752,7 +753,7 @@ func (hs *clientHandshakeStateTLS13) abortIfRequired() error {
 	if c.ech.offered && !c.ech.accepted {
 		// If ECH was rejected, then abort the handshake.
 		c.sendAlert(alertECHRequired)
-		return errors.New("tls: ech: rejected")
+		return errors.New("tls: ech: rejected " + string(base64.StdEncoding.EncodeToString(c.ech.retryConfigs)))
```
chris-wood, line 756: Why do we need to cast? `EncodeToString` returns a `string`, no?

chris-wood (approved):
LGTM with a few nits.

src/crypto/tls/ech.go
cjpatton: Done, here and elsewhere in this file.

src/crypto/tls/ech.go
cjpatton: Isn't that asserted above?
```
	if !s.Empty() &&
		(!s.ReadUint16LengthPrefixed(&extensions) || !s.Empty()) { // <--- ASSERTED AFTER "||"
		return nil, false
	}
```

src/crypto/tls/handshake_client_tls13.go
cjpatton: Oops, that's right. But I also didn't mean to commit this! Reverted.

src/crypto/tls/ech.go
chris-wood: Oh, yeah. It's just not easy to read (for me). I'd recommend moving that to the following line to make it clear.

chris-wood (approved):

src/crypto/tls/ech.go
cjpatton: Done.

Issue #32, opened 2020-09-24
Implement HPKE in CIRCL


cjpatton:
ECH requires HPKE. Currently the implementation that will be used is githbub.com/cisco/go-hpke (see #30), which in turn depends on github.com/cloudflare/circl for PQ KEMs. Let's add HPKE to circl to close this loop. For now we will only need (X25519, HKDF-SHA256, AES128-GCM).

cjpatton:
@armfazh brought up this idea here: https://github.com/cloudflare/go/pull/30#issuecomment-697758842.

cjpatton:
PR for HPKE In CIRCL: https://github.com/cloudflare/circl/pull/170

cjpatton:
Closed because work is in progress elsewhere.

PR #40, opened 2020-11-23
Vendor CIRCL properly

cjpatton:
Partially addresses #32.
Requires https://github.com/cloudflare/circl/pull/192.

As we implement more experiments, we are likely to depend on CIRCL more and more. Currently, updating CIRCL requires running a script that copies CIRCL into `src/circl` without the help of the go tools. The goal of this PR is to make updating CIRCL more ergonomic:

- Removes the `cf-copy-from-cricle.py` script and reverts corresponding changes to `src/go/build/deps_test.go`
- Removes `srcl/circl`.
- Adds CIRCL to `src/vendor/` and modifies `src/go/build/deps_test.go` to allow standard library packages to depend on it.

Now dependencies can be updated with `go get` and `go vendor` as usual.

However, before this can be merged, one more issue remains. `all.bash` works, except for one failure I can't fully explain:
```
--- FAIL: TestDependencyVersionsConsistent (0.00s)
    moddeps_test.go:217: Modules within GOROOT require different versions of golang.org/x/sys.
    moddeps_test.go:229: cmd	requires v0.0.0-20200501145240-bc7a7d42d5c3
    moddeps_test.go:229: std	requires v0.0.0-20200602225109-6fdc65e7d980
FAIL
```
This seems to be happening because the `cmd` module uses my system's version of Go and `std` uses this Go fork.


src/go.mod
```diff
@@ -3,8 +3,8 @@ module std
 go 1.15
 
 require (
+	github.com/cloudflare/circl v1.0.1-0.20201123155603-cd1efefc7f5f
```
Lekensteyn, line 6: Not sure what the effect is of including dependencies here on applications using the standard Go library. In any case this will cause conflicts during major updates that update this file.

Lekensteyn:
> However, before this can be merged, one more issue remains. `all.bash` works, except for one failure I can't fully explain:
> 
> ```
> --- FAIL: TestDependencyVersionsConsistent (0.00s)
>     moddeps_test.go:217: Modules within GOROOT require different versions of golang.org/x/sys.
>     moddeps_test.go:229: cmd	requires v0.0.0-20200501145240-bc7a7d42d5c3
>     moddeps_test.go:229: std	requires v0.0.0-20200602225109-6fdc65e7d980
> FAIL
> ```
> 
> This seems to be happening because the `cmd` module uses my system's version of Go and `std` uses this Go fork.

The reason is that the cmd and std packages have their own go.mod files and their functions must remain in sync. See also https://github.com/golang/go/issues/30241

@bwesterb surely has some additional insights why the circl library is vendored as it does now.

bwesterb:
> @bwesterb surely has some additional insights why the circl library is vendored as it does now.

One of the reasons is that I didn't get `golang.org/x/sys/cpu` to work which in the standard library is called `internal/cpu`.

I prefer the approach of this PR (if we get it to work, of course.)

src/go.mod
bwesterb: Yes, this should also be checked.

cjpatton:
Based on the issue that @Lekensteyn pointed to, it seems like this change isn't such a good idea. What I want is for github.com/cloudflare/circl to be a dependency of the standard library, just as golang.org/x is a dependency of the standard library. But github.com/cloudflare/circl depends, in turn, on golang.org/x; and if the version of golang.org/x required by github.com/cloudflare/circl is not the same as the standard library, we end up with a conflict. It seems like the Go standard library wants to avoid the conditions that lead to this conflict.

armfazh:
Using `replace` directive may help
 https://golang.org/ref/mod#go-mod-file-replace

cjpatton:
Where would the directive go?

armfazh:
> Where would the directive go?

it could be appended to a go.mod file to make circl point the same golang.org/x, it is matter of testing if it works.

cjpatton:
Right, my question is which go.mod? github.com/circl/go.mod? github.com/cloudfalre/go/src/go.mod? github.com/cloudflare/go/cmd/go.mod?

armfazh:
> Right, my question is which go.mod? github.com/circl/go.mod? github.com/cloudfalre/go/src/go.mod? github.com/cloudflare/go/cmd/go.mod?

I did this hacky patch, so we can move forward 

https://github.com/cloudflare/go/pull/41

cjpatton:
@Lekensteyn and I just spoke about this, and here's where we landed. The main problem with this strategy is that updating CIRCL might change the standard libraries dependencies in golang.org/x. They're being upgraded in this particular instance, but in general they could be downgraded. Either way, changing the dependencies might have side-effects that we don't understand. This is especially problematic if we want to keep our fork in-sync with Go releases.

The advantage of the status quo is that it effectively defines CIRCL as being part of the standard library. I'll open another PR that adds support for CIRCL's HPKE implementation (https://github.com/cloudflare/circl/pull/170) using the existing script and see if I can get that to work.

bwesterb:
Another option is to copy `x/sys/cpu` into CIRCL (like we do with `x/crypto/sha3` so that Circl has no dependencies at all.

cjpatton:
Closed in favor of #42.

Issue #27, opened 2020-08-03
Should we add a CI?


claucece:
Do you think it will be worth adding a CI here? For testing on different archs... @bwesterb

Lekensteyn:
It would be helpful to have to ensure that PRs do not cause regressions elsewhere, so 👍 for CI.

GitHub Actions supports Windows, Linux and macOS (https://docs.github.com/en/actions/getting-started-with-github-actions/about-github-actions). It also claims to support ARM, but I could not quickly find documentation about that.

circl does use GitHub Actions for arm64, an example can be found here: https://github.com/cloudflare/circl/blob/master/.github/workflows/ci-actions.yml

claucece:
Yes, I think using  GitHub Actions might be the right way. I'll integrate it ;)

thomwiggers:
For ARM support, you need to provide self-hosted runners afaik.

claucece:
Thank you @Lekensteyn

PR #42, opened 2020-11-25
Update CIRCL

### LINKED ISSUE #43: Improve CIRCL upgrade procedure
One of the main differences between cfgo (there I said) and upstream go is that our fork includes CIRCL in the standard library. To updated CIRCL, we run a script called `cf-update-circl.py` that:
1. clones the master branch of github.com/cloudflare/circl;
1. copies its  contents into `src/circl`;
1. prunes non-package files and directories (e.g., `.git`, ASM generators, etc.); and
1. fixes imports (`github.com/cloudflare/circl/kem` -> `circl/kem` and so on, `golang.org/x/sys/cpu` to `internal/cpu`).

This process has three problems that I can see:
1. The script requires occasional maintenance, as changes to upstream CIRCL can break the build. This is inevitable, of course, but CIRCL is currently changing much more quickly than cfgo's requirements for CIRCL change. Instead of copying-then-pruning, we want to make sure that we only build the dependencies we need.
1. Import fix-ups are done by a simple search-and-replace, which is a bit icky because, in theory, it could change the code.
1. AFAICT there's no provenance of which version of CIRCL is running. This would be useful for debugging purposes.

There are two ways we can try to remedy this. The first (PR #40) is to use go modules to vendor CIRCL properly. This works pretty well, except that resolving dependencies shared by the standard library and CIRCL requires changing the standard library's dependencies. As we don't know what the side effects of these changes might be, we ought to consider these changes as "unsafe" in general.

With that option off the table, our best option is to improve the existing script. That work is tracked by PR #42.

cjpatton:
Closes #43.

1. Upgrades CIRCL.
2. ~Removes `cf-copy-from-circle.py` and replaces it with `cf-update-circl.py`. Instead of copying everything and removing unused packages, the new script defines a manifest of packages that are needed for the standard library and copies them. This reduces the risk of changes to CIRCL breaking the build of the Go standard library.~


cf-update-circl.py
```diff
@@ -0,0 +1,195 @@
...
+    with open(fn, 'wb') as f:
+        f.write(header)
+        f.write(s)
+
+pkg_file_header = '''
+// THIS CODE WAS GENERATED BY cf-update-circl.py.
+//
+// It was checked out from "{0}"
+// on branch "{1}". Latest commit:
+//'''
```
Lekensteyn, line 99: Let's not add it to every individual file. This info can be read from the git history for the Go repo. If it is really important to have it in the source tree, I would suggest creating one file with this metadata, e.g. `src/circle/about.txt`

cjpatton: Just curious, what do you see as the downside of adding it to each file? Big diffs?

cjpatton: (Done.)

armfazh:
cc; @bwesterb

cf-update-circl.py
Lekensteyn: Big diffs to a lesser extent, I was more thinking about comparing two files or directories (src/circl/ vs circl upstream).

cf-update-circl.py
```diff
@@ -0,0 +1,195 @@
...
+# standard library.
+def handle_pkg_file(fn):
+    with open(fn, 'rb') as f:
+        s = f.read()
+
+    # Fix Go imports.
+    if fn.endswith('.go'):
+        s = s.replace(
+            b'github.com/cloudflare/circl',
+            b'circl',
```
bwesterb, line 84: `circl_name`

cf-update-circl.py
```diff
@@ -0,0 +1,195 @@
...
+    return ok
+
+# Modifies file fn by replacing imports with the corresponding package in the
+# standard library.
+def handle_pkg_file(fn):
+    with open(fn, 'rb') as f:
+        s = f.read()
+
+    # Fix Go imports.
+    if fn.endswith('.go'):
```
bwesterb, line 81: Reading and writing can be moved into the if statement.

cf-update-circl.py
```diff
@@ -0,0 +1,195 @@
...
+circl_name = 'circl'
+
+# The git repository form which CIRCL will be cloned.
+circl_repo = 'https://github.com/cloudflare/circl'
+
+# The git branch to checkout.
+circl_branch = 'master'
+
+# The set of CIRCL packages that the standard library depends on.
+circl_pkg_dir = [
```
bwesterb, line 19: Instead of keeping a list of files/directories to remove, you've switched to keeping a list of files/directories to keep.  My guess is that this will be more work during updates.

bwesterb (approved):

cf-update-circl.py
cjpatton: Nice catch!

cf-update-circl.py
cjpatton: Done.

cf-update-circl.py
cjpatton: I think that's true when adding a new CIRCL dependency. However, the motivation for this change is that updating CIRCL --- whether you're adding a new feature or not --- tends to break the build here, sometimes requiring upstream changes. By copying only what we need, we minimize this risk. My bet is that overtime, CIRCL will change more rapidly than cfgo's CIRCL dependencies change. This has been true lately. If this holds true in the future, then I think this will be an improvement.

cf-update-circl.py
```diff
@@ -0,0 +1,193 @@
...
+# Modifies file fn by replacing imports with the corresponding package in the
+# standard library.
+def handle_pkg_file(fn):
+    # Fix Go imports.
+    if fn.endswith('.go'):
+        with open(fn, 'rb') as f:
+            s = f.read()
+        s = s.replace(
+            b'github.com/cloudflare/circl',
+            bytes(circl_name, 'ascii'),
```
Lekensteyn, line 83: All .go files are expected to be UTF-8, so you might as well use `open(fn, 'r')` and `open(fn, 'w')` and drop the `b` prefixes.

If you do not, usually in Python one uses `circl_name.encode('utf8')` instead of `bytes(circl_name, 'utf8')`.

cf-update-circl.py
```diff
@@ -0,0 +1,193 @@
...
+    for (f, v) in misc.items():
+        if not v:
+            print(f'WARNING: file "{f}" not found in CIRCL repo', file=sys.stderr)
+
+    print(f"Creating README ...")
+    with open(os.path.join(src_circl, 'README'), 'w') as f:
+        readme = subprocess.check_output(['git', 'log',
+            '-n1',
+            '--format=fuller',
+        ], cwd=os.path.join(tmp, circl_name))
```
Lekensteyn, line 178: To avoid the next line (`readme = str(readme, 'ascii')`) consider:
```suggestion
        ], cwd=os.path.join(tmp, circl_name), universal_newlines=True)
```

You are already using f-strings which require Python 3.6. If you can use Python 3.7, you can use `text=True` instead of `universal_newlines=True` for the same effect.

cf-update-circl.py
```diff
@@ -0,0 +1,193 @@
...
+
+        # Copy miscellaneous files.
+        for f in files:
+            k = os.path.join(d, f)
+            if k in misc.keys():
+                misc[k] = True
+                subprocess.check_call(['cp',
+                    os.path.join(subdir, f),
+                    os.path.join(src, d, f),
+                ])
```
Lekensteyn, line 159: It might be worth printing files that were ignored in case they were important?

Lekensteyn (commented):
I'm inclined to prefer the explicit removal approach since that requires less hardcoding of paths. With an explicit list of entries to keep, it becomes more likely for the build to fail on circl changes.

What would you prefer @bwesterb ?

Lekensteyn:
Also please have a look at addressing #27 first, ideally we have:
- One job for amd64
- One job for arm64 (if possible)
- One job that runs this update script

With the last job, we would know when the script is about to fail.

cf-update-circl.py
cjpatton: Dropped `b` prefixes and changed `'ascii'` to `'utf8'` below.

cf-update-circl.py
cjpatton: Done.

armfazh:
This approach looks nice. Choosing either inclusion or exclusion of files might require manual operation when circl updates. I have no preference between them. 

Another alternative is including everything from circl and let the compiler links what is needed (however, this option needs more effort to be accomplished).

cf-update-circl.py
cjpatton: Done.

bwesterb:
> I'm inclined to prefer the explicit removal approach since that requires less hardcoding of paths. With an explicit list of entries to keep, it becomes more likely for the build to fail on circl changes.
> 
> What would you prefer @bwesterb ?

@Lekensteyn I prefer listing the entries to remove if I were to maintain this, but I'm not.

cjpatton:
To @Lekensteyn's point, whether we do inclusion or exclusion requires some amount of maintenance. Which requires more maintenance depends on which changes more often: CIRCL, or cfgo's dependencies on CIRCL. The advantage of inclusion is that if CIRCL changes faster than cfgo's dependencies on CRICL, then less maintenance is required. On the other hand, if CIRCL changes slower than cfgo's dependencies, then we're better off with exclusion.

Recent experience upgrading CIRCL bares out that CIRCL changes faster than cfgo's dependencies on it, hence my support for inclusion.

cjpatton:
Also, note that the list of things currently excluded is longer than the list of things currently included.

cjpatton:
It seems like there's a weak preference for the status quo. so I'll revert the script change and merge the CIRCL update.

PR #33, opened 2020-10-01
Implement KEMTLS based on Delegated Credentials

thomwiggers:
This PR implements [KEMTLS](https://ia.cr/2020/534) based on @claucece's DC branch (#28). There are a few minor fixes included in the DC stuff, surrounding the negotiation of DC supported algorithms.

This branch is targeted at the `cf` because that makes this PR reviewable: otherwise all of `circl` is included on top of my changes to the DC implementation and `crypto/tls` in general.

The following is currently implemented
* [x] KEMTLS
* [x] `crypto/kem` ergonomic KEM api
  * This package implements:
    * Kem25519: a KEM-variant of X25519. Useful for testing, but should be deleted before merge.
    * SIKEp434 provided via `circl`
    * Kyber512 provided via `circl`
* [ ] KEMTLS client authentication

## New TLS options
New supported signature algorithms (for DC):
  * [x] KEMTLSwithSIKEp434
  * [x] KEMTLSwithKyber512

New supported keyshare types (these can be negotiated independently in TLS 1.3):
* [x] SIKEp434
* [x] Kyber512

## Testing
* [ ] Unit tests
   * I've currently tested it by hard-coding the type of DC that the DC tests use. Unfortunately, those tests still need to be refactored a bit before it's easy to add more types of supported DC — but I'm sure that was on @claucece's list :smiley: 

## Stuff to do still:
* Expand the number of KEMs based on newer versions of Circl (~Kyber, hopefully, soon~)

PR #44, opened 2020-12-09
Implement KEM TLS 

claucece:
From #33 

cc./ @thomwiggers @bwesterb @chris-wood


claucece:
@thomwiggers I'm going one by one in the commits and finding things ;) So still a WIP

src/crypto/kem/kem.go
```diff
@@ -0,0 +1,209 @@
...
+)
+
+// ID identifies each flavor of KEM.
+type ID uint16
+
+const (
+	// KEM25519 is X25519 as a KEM. Not quantum-safe.
+	KEM25519 ID = 0x01fb
+	// CSIDH is a post-quantum NIKE
+	CSIDH ID = 0x01fc
```
thomwiggers, line 21:
```suggestion
```
I probably initially included this because that's the "first" KEM supported by the Rustls stack; but it is not used elsewhere and breaks the validation logic with `minKEM`, `maxKEM`.

thomwiggers: Relatedly: the kems need to be renumbered.

thomwiggers:
https://github.com/cloudflare/go/compare/cf-delegated-credentials...cf-pq-kem-tls is useful to compare with the DC PR #28 directly.

src/crypto/kem/kem.go
claucece: Yeah.. I'll change them ;)

claucece: Should be better now.

src/crypto/tls/handshake_client_test.go
```diff
@@ -527,16 +527,6 @@ func runClientTestTLS13(t *testing.T, template *clientTest) {
...
-func TestHandshakeClientRSARC4(t *testing.T) {
-	test := &clientTest{
-		name: "RSA-RC4",
-		args: []string{"-cipher", "RC4-SHA"},
-	}
-	runClientTestTLS10(t, test)
-	runClientTestTLS11(t, test)
-	runClientTestTLS12(t, test)
-}
-
```
thomwiggers, line 539: Did you compile OpenSSL 1.1 with weak crypto turned on?

thomwiggers: compiling with `no-shared` might also help https://github.com/golang/go/pull/41630

claucece: Yeah.. with weak crypto. I'll try the no-shared...

src/crypto/kem/kem.go
```diff
@@ -0,0 +1,207 @@
...
+			return nil, nil, err
+		}
+
+		return ss, ct, nil
+	case KEM25519:
+		privateKey := make([]byte, curve25519.ScalarSize)
+		if _, err := io.ReadFull(rand, privateKey); err != nil {
+			return nil, nil, err
+		}
+		ciphertext, err := curve25519.X25519(privateKey, curve25519.Basepoint)
```
armfazh, line 133: we can use circl here too.

src/crypto/tls/auth.go
```diff
@@ -203,54 +204,65 @@ var rsaSignatureSchemes = []struct {
...
-			sigAlgs = []SignatureScheme{
-				ECDSAWithP256AndSHA256,
-				ECDSAWithP384AndSHA384,
-				ECDSAWithP521AndSHA512,
-				ECDSAWithSHA1,
-			}
-			break
+	var priv crypto.Signer
+
+	kemPriv, ok := cert.PrivateKey.(*kem.PrivateKey)
```
claucece, line 210: This should be generalised to crypto.Signer as well.

thomwiggers:
Looks decent to me on the whole :+1:

src/crypto/kem/kem.go
claucece: I'll add this for future, after RWC.

claucece:
Closing this in favour of the kemtls branch.

thomwiggers:
Do you mean that it'll just live in this separate branch from now on? (and which of the two branches)

claucece:
@thomwiggers  The branch that will be used is `kemtls`. I'm not sure if this will make it to `cf` branch, so I'll have to see what consensus is around this, but that might take a while. I'll ask next week.

PR #45, opened 2020-12-10
Add experimental handshake timings API

chris-wood:
The basic idea here is as follows:

Extend the experimental event (`EXP_Event`) interface to cover "handshake timing info" events, which tracks relevant timestamps for important events in the TLS 1.3 state machine. Applications that wish to use these timestamps to track metrics, such as total handshake latency from CH...SFIN.

For example, applications might wire up an event handler to collect this timing information and use it as needed:

~~~go
// ... modified from metrics_test.go ...

func (r *testTimingInfo) eventHandler(event EXP_Event) {
	switch e := event.(type) {
	case EXP_EventTLS13ServerHandshakeTimingInfo:
		// use e's server timestamps as needed
	case EXP_EventTLS13ClientHandshakeTimingInfo:
		// use e's client timestamps as needed
	}
}

config := tls.Config{
   EXP_EventHandler: eventHandler
} 

// run TLS connection with `config`
}
~~~

### DIFF
--- a/src/crypto/tls/handshake_client.go
+++ b/src/crypto/tls/handshake_client.go
@@ -140,6 +140,8 @@ func (c *Conn) clientHandshake() (err error) {
 		c.config = defaultConfig()
 	}
 
+	handshakeTimings := createTLS13ClientHandshakeTimingInfo(c.config.Time)
+
 	// This may be a renegotiation handshake, in which case some fields
 	// need to be reset.
 	c.didResume = false
@@ -186,6 +188,8 @@ func (c *Conn) clientHandshake() (err error) {
 		return err
 	}
 
+	handshakeTimings.WriteClientHello = handshakeTimings.elapsedTime()
+
 	msg, err := c.readHandshake()
 	if err != nil {
 		return err
@@ -215,15 +219,16 @@ func (c *Conn) clientHandshake() (err error) {
 
 	if c.vers == VersionTLS13 {
 		hs := &clientHandshakeStateTLS13{
-			c:           c,
-			serverHello: serverHello,
-			hello:       hello,
-			helloInner:  helloInner,
-			helloBase:   helloBase,
-			ecdheParams: ecdheParams,
-			session:     session,
-			earlySecret: earlySecret,
-			binderKey:   binderKey,
+			c:                c,
+			serverHello:      serverHello,
+			hello:            hello,
+			helloInner:       helloInner,
+			helloBase:        helloBase,
+			ecdheParams:      ecdheParams,
+			session:          session,
+			earlySecret:      earlySecret,
+			binderKey:        binderKey,
+			handshakeTimings: handshakeTimings,
 		}
 
 		// In TLS 1.3, session tickets are delivered after the handshake.
--- a/src/crypto/tls/handshake_client_tls13.go
+++ b/src/crypto/tls/handshake_client_tls13.go
@@ -17,7 +17,7 @@ import (
 )
 
 // EXP_EventTLS13ClientHandshakeTimingInfo carries intra-stack time durations
-// for state machine changes. It can be used for tracking metrics during a
+// for TLS 1.3 state machine changes. It can be used for tracking metrics during a
 // connection. Some durations may be sensitive, such as the amount of time to
 // process a particular handshake message, so this event should only be used
 // for experimental purposes.
@@ -27,7 +27,7 @@ type EXP_EventTLS13ClientHandshakeTimingInfo struct {
 	timer                   func() time.Time
 	start                   time.Time
 	WriteClientHello        time.Duration
-	ReadServerHello         time.Duration
+	ProcessServerHello      time.Duration
 	ReadEncryptedExtensions time.Duration
 	ReadCertificate         time.Duration
 	ReadCertificateVerify   time.Duration
@@ -389,7 +389,7 @@ func (hs *clientHandshakeStateTLS13) processServerHello() error {
 	c := hs.c
 
 	defer func() {
-		hs.handshakeTimings.ReadServerHello = hs.handshakeTimings.elapsedTime()
+		hs.handshakeTimings.ProcessServerHello = hs.handshakeTimings.elapsedTime()
 	}()
 
 	if bytes.Equal(hs.serverHello.random, helloRetryRequestRandom) {
--- a/src/crypto/tls/handshake_client_tls13.go
+++ b/src/crypto/tls/handshake_client_tls13.go
@@ -16,9 +16,11 @@ import (
 	"time"
 )
 
-// EXP_EventTLS13ClientHandshakeTimingInfo carries intra-stack durations for
-// state machine changes. It can be used for tracking metrics during a
-// connection.
+// EXP_EventTLS13ClientHandshakeTimingInfo carries intra-stack time durations
+// for state machine changes. It can be used for tracking metrics during a
+// connection. Some durations may be sensitive, such as the amount of time to
+// process a particular handshake message, so this event should only be used
+// for experimental purposes.
 //
 // NOTE: This API is EXPERIMENTAL and subject to change.
 type EXP_EventTLS13ClientHandshakeTimingInfo struct {
--- a/src/crypto/tls/handshake_client_tls13.go
+++ b/src/crypto/tls/handshake_client_tls13.go
@@ -16,6 +16,46 @@ import (
 	"time"
 )
 
+// EXP_EventTLS13ClientHandshakeTimingInfo carries intra-stack durations for
+// state machine changes. It can be used for tracking metrics during a
+// connection.
+//
+// NOTE: This API is EXPERIMENTAL and subject to change.
+type EXP_EventTLS13ClientHandshakeTimingInfo struct {
+	timer                   func() time.Time
+	start                   time.Time
+	WriteClientHello        time.Duration
+	ReadServerHello         time.Duration
+	ReadEncryptedExtensions time.Duration
+	ReadCertificate         time.Duration
+	ReadCertificateVerify   time.Duration
+	ReadServerFinished      time.Duration
+	WriteCertificate        time.Duration
+	WriteCertificateVerify  time.Duration
+	WriteClientFinished     time.Duration
+}
+
+// Name is required by the EXP_Event interface.
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) Name() string {
+	return "TLS13ClientHandshakeTimingInfo"
+}
+
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) elapsedTime() time.Duration {
+	return time.Now().Sub(e.start)
+}
+
+func createTLS13ClientHandshakeTimingInfo(timerFunc func() time.Time) EXP_EventTLS13ClientHandshakeTimingInfo {
+	timer := time.Now
+	if timerFunc != nil {
+		timer = timerFunc
+	}
+
+	return EXP_EventTLS13ClientHandshakeTimingInfo{
+		timer: timer,
+		start: timer(),
+	}
+}
+
 type clientHandshakeStateTLS13 struct {
 	c           *Conn
 	serverHello *serverHelloMsg
@@ -36,6 +76,8 @@ type clientHandshakeStateTLS13 struct {
 	transcriptInner hash.Hash
 	masterSecret    []byte
 	trafficSecret   []byte // client_application_traffic_secret_0
+
+	handshakeTimings EXP_EventTLS13ClientHandshakeTimingInfo
 }
 
 // handshake requires hs.c, hs.hello, hs.serverHello, hs.ecdheParams, and,
@@ -132,6 +174,7 @@ func (hs *clientHandshakeStateTLS13) handshake() error {
 		return err
 	}
 
+	c.handleEvent(hs.handshakeTimings)
 	atomic.StoreUint32(&c.handshakeStatus, 1)
 
 	return nil
@@ -343,6 +386,10 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
 func (hs *clientHandshakeStateTLS13) processServerHello() error {
 	c := hs.c
 
+	defer func() {
+		hs.handshakeTimings.ReadServerHello = hs.handshakeTimings.elapsedTime()
+	}()
+
 	if bytes.Equal(hs.serverHello.random, helloRetryRequestRandom) {
 		c.sendAlert(alertUnexpectedMessage)
 		return errors.New("tls: server sent two HelloRetryRequest messages")
@@ -469,6 +516,8 @@ func (hs *clientHandshakeStateTLS13) readServerParameters() error {
 		}
 	}
 
+	hs.handshakeTimings.ReadEncryptedExtensions = hs.handshakeTimings.elapsedTime()
+
 	return nil
 }
 
@@ -518,6 +567,8 @@ func (hs *clientHandshakeStateTLS13) readServerCertificate() error {
 	}
 	hs.transcript.Write(certMsg.marshal())
 
+	hs.handshakeTimings.ReadCertificate = hs.handshakeTimings.elapsedTime()
+
 	c.scts = certMsg.certificate.SignedCertificateTimestamps
 	c.ocspResponse = certMsg.certificate.OCSPStaple
 
@@ -558,6 +609,8 @@ func (hs *clientHandshakeStateTLS13) readServerCertificate() error {
 
 	hs.transcript.Write(certVerify.marshal())
 
+	hs.handshakeTimings.ReadCertificateVerify = hs.handshakeTimings.elapsedTime()
+
 	return nil
 }
 
@@ -574,6 +627,8 @@ func (hs *clientHandshakeStateTLS13) readServerFinished() error {
 		c.sendAlert(alertUnexpectedMessage)
 		return unexpectedMessageError(finished, msg)
 	}
+
+	hs.handshakeTimings.ReadServerFinished = hs.handshakeTimings.elapsedTime()
 
 	expectedMAC := hs.suite.finishedHash(c.in.trafficSecret, hs.transcript)
 	if !hmac.Equal(expectedMAC, finished.verifyData) {
@@ -634,6 +689,8 @@ func (hs *clientHandshakeStateTLS13) sendClientCertificate() error {
 		return err
 	}
 
+	hs.handshakeTimings.WriteCertificate = hs.handshakeTimings.elapsedTime()
+
 	// If we sent an empty certificate message, skip the CertificateVerify.
 	if len(cert.Certificate) == 0 {
 		return nil
@@ -672,6 +729,8 @@ func (hs *clientHandshakeStateTLS13) sendClientCertificate() error {
 		return err
 	}
 
+	hs.handshakeTimings.WriteCertificateVerify = hs.handshakeTimings.elapsedTime()
+
 	return nil
 }
 
@@ -686,6 +745,8 @@ func (hs *clientHandshakeStateTLS13) sendClientFinished() error {
 	if _, err := c.writeRecord(recordTypeHandshake, finished.marshal()); err != nil {
 		return err
 	}
+
+	hs.handshakeTimings.WriteClientFinished = hs.handshakeTimings.elapsedTime()
 
 	c.out.setTrafficSecret(hs.suite, hs.trafficSecret)
 
--- a/src/crypto/tls/handshake_server_tls13.go
+++ b/src/crypto/tls/handshake_server_tls13.go
@@ -22,9 +22,11 @@ import (
 // messages cause too much work in session ticket decryption attempts.
 const maxClientPSKIdentities = 5
 
-// EXP_EventTLS13ServerHandshakeTimingInfo carries intra-stack durations for
-// state machine changes. It can be used for tracking metrics during a
-// connection.
+// EXP_EventTLS13ServerHandshakeTimingInfo carries intra-stack time durations
+// for state machine changes. It can be used for tracking metrics during a
+// connection. Some durations may be sensitive, such as the amount of time to
+// process a particular handshake message, so this event should only be used
+// for experimental purposes.
 //
 // NOTE: This API is EXPERIMENTAL and subject to change.
 type EXP_EventTLS13ServerHandshakeTimingInfo struct {
--- a/src/crypto/tls/handshake_server_tls13.go
+++ b/src/crypto/tls/handshake_server_tls13.go
@@ -22,6 +22,46 @@ import (
 // messages cause too much work in session ticket decryption attempts.
 const maxClientPSKIdentities = 5
 
+// EXP_EventTLS13ServerHandshakeTimingInfo carries intra-stack durations for
+// state machine changes. It can be used for tracking metrics during a
+// connection.
+//
+// NOTE: This API is EXPERIMENTAL and subject to change.
+type EXP_EventTLS13ServerHandshakeTimingInfo struct {
+	timer                    func() time.Time
+	start                    time.Time
+	ReadClientHello          time.Duration
+	WriteServerHello         time.Duration
+	WriteEncryptedExtensions time.Duration
+	WriteCertificate         time.Duration
+	WriteCertificateVerify   time.Duration
+	WriteServerFinished      time.Duration
+	ReadCertificate          time.Duration
+	ReadCertificateVerify    time.Duration
+	ReadClientFinished       time.Duration
+}
+
+// Name is required by the EXP_Event interface.
+func (e EXP_EventTLS13ServerHandshakeTimingInfo) Name() string {
+	return "TLS13ServerHandshakeTimingInfo"
+}
+
+func (e EXP_EventTLS13ServerHandshakeTimingInfo) elapsedTime() time.Duration {
+	return e.timer().Sub(e.start)
+}
+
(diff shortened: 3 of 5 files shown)


claucece:
I really like this idea. Thank you @chris-wood !

cjpatton:
This is a great idea, but I think I would go about it differently. First, instead of recording time stamps for each state change, I would record the duration from the start of the handshake to the state change. This avoids surfacing timestamps, which are potentially privacy sensitive.

Second, why not just make this an event type?
```
type EXP_EventHandshakeTimingInfo struct {
   TimeToState1 time.Duration
   TimeToState2 time.Duration
   ....
   TimeToStateN time.Duration
}
```
The time intervals would be recorded at each state change. The event itself would be emitted by the server once, after the handshake is completed. The advantage of this approach is that the event handler doesn't have to be stateful; it just records the metrics as soon as it receives them. Alternatively, we could extend the `EXP_Event` type so that it exposes the time elapsed since the handshake began.
```
type EXP_Event interface {
   Name() string
   ElapsedTime() time.Duration     // <--- new interface method
}
```
Then we would define an event for each state change we're interested in.

Thirdly, I'm not sure I see a benefit of having events specify an opaque value.

chris-wood:
@cjpatton this should be mostly ready for review. I'm sure we can move timestamp locations around, if needed, but that shouldn't block the main API.

src/crypto/tls/handshake_client.go
```diff
@@ -23,6 +23,31 @@ import (
 	circlSign "circl/sign"
 )
 
+// EXP_EventServerHandshakeTimingInfo is carries intra-stack durations for
```
chris-wood, line 26:
```suggestion
// EXP_EventClientHandshakeTimingInfo is carries intra-stack durations for
```

src/crypto/tls/handshake_client.go
```diff
@@ -23,6 +23,31 @@ import (
 	circlSign "circl/sign"
 )
 
+// EXP_EventClientHandshakeTimingInfo is carries intra-stack durations for
```
cjpatton, line 26:
```suggestion
// EXP_EventClientHandshakeTimingInfo carries intra-stack durations for
```

cjpatton, line 26: This event will only be emitted during a TLS 1.3 handshake, right? I suggest adding "TLS13" to the name.

src/crypto/tls/handshake_client.go
```diff
@@ -186,6 +213,12 @@ func (c *Conn) clientHandshake() (err error) {
 		return err
 	}
 
+	handshakeTimings := &EXP_EventClientHandshakeTimingInfo{
+		start: time.Now(),
```
cjpatton, line 217: use `c.config.Time()` instead of `time.Now()`.

src/crypto/tls/handshake_client.go
```diff
@@ -31,6 +56,8 @@ type clientHandshakeState struct {
 	finishedHash finishedHash
 	masterSecret []byte
 	session      *ClientSessionState
+
+	handshakeTimings *EXP_EventClientHandshakeTimingInfo
```
cjpatton, line 60: Not needed for TLS 1.2 and below, so I would drop this.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -36,6 +36,8 @@ type clientHandshakeStateTLS13 struct {
 	transcriptInner hash.Hash
 	masterSecret    []byte
 	trafficSecret   []byte // client_application_traffic_secret_0
+
+	handshakeTimings *EXP_EventClientHandshakeTimingInfo
```
cjpatton, line 40: I would put this on the stack rather than the heap. (I.e., don't make this a pointer.)

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -132,6 +134,7 @@ func (hs *clientHandshakeStateTLS13) handshake() error {
 		return err
 	}
 
+	c.handleEvent(*hs.handshakeTimings)
```
cjpatton, line 137: I would add this callback to the end of `sendClientFinished()`.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -343,6 +346,8 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
 func (hs *clientHandshakeStateTLS13) processServerHello() error {
 	c := hs.c
 
+	hs.handshakeTimings.ReadServerHello = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 349: Should this go after the (HRR and the) SH has been processed?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -450,6 +455,9 @@ func (hs *clientHandshakeStateTLS13) readServerParameters() error {
 		c.sendAlert(alertUnexpectedMessage)
 		return unexpectedMessageError(encryptedExtensions, msg)
 	}
+
+	hs.handshakeTimings.ReadEncryptedExtensions = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 459: I'd bump this to the end of the function so that this duration accounts for the time it takes to process the inputs (and not just read them). Here and below.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -22,6 +22,45 @@ import (
 // messages cause too much work in session ticket decryption attempts.
 const maxClientPSKIdentities = 5
 
+// EXP_EventServerHandshakeTimingInfo is carries intra-stack durations for
```
cjpatton, line 25: It's odd that this is defined in `handshake_server_tls13.go` but `EXP_EventClientHandshakeTimingInfo` is defined in `handshake_client.go`.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -22,6 +22,45 @@ import (
...
+// Name is required by the EXP_Event interface.
+func (e EXP_EventServerHandshakeTimingInfo) Name() string {
+	return "ServerHandshakeTimingInfo"
+}
+
+func (e EXP_EventServerHandshakeTimingInfo) elapsedTime() time.Duration {
+	return time.Now().Sub(e.start)
+}
+
+const (
```
cjpatton, line 50: These constants and `infoNames` below can be dropped, as they're not used anywhere else.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -38,6 +77,8 @@ type serverHandshakeStateTLS13 struct {
 	trafficSecret   []byte // client_application_traffic_secret_0
 	transcript      hash.Hash
 	clientFinished  []byte
+
+	handshakeTimings *EXP_EventServerHandshakeTimingInfo
```
cjpatton, line 81: As above, no need to use a pointer for this.

cjpatton (changes requested):
High level comments:
1. Open question: What do we do if the client or server aborts early?
1. Right now the "read" state change timings are recorded after the bits have been read on the wire, but before the bits have been parsed and processed. Maybe we should change "read" state changes to "processed" (e.g., "ReadServerHello" -> "ServerHelloConsumed"). 
1. We should write some basic unit tests to make sure the duration is monotonically increasing. This will help make sure we don't forget to record certain metrics.

src/crypto/tls/tls_test.go
```diff
@@ -1071,7 +1071,7 @@ func TestConnectionState(t *testing.T) {
 			name = "TLSv13"
```
cjpatton, line 1071: Instead of modifying an existing test, I think we should create a new file for validating events. E.g., `tls_cf_test.go`? I'm pretty familiar with how to set up a basic handshake test, if you'd like i can push a test case here.

src/crypto/tls/handshake_client_tls13.go
chris-wood: I kept it here because KEMTLS (or other things) might want to record more events after sending CFIN. (KEMTLS, for example, sends CFIN before SFIN!) Since the time isn't affected, I think it's fine here?

src/crypto/tls/handshake_client_tls13.go
chris-wood: Hmm... presumably after SH.

src/crypto/tls/handshake_client_tls13.go
chris-wood: Good call.

src/crypto/tls/handshake_server_tls13.go
chris-wood: Yeah, that's leftover from including this in the 1.2 CH struct. I'll remove it from there, and then move this to the 1.3-specific file.

src/crypto/tls/handshake_server_tls13.go
chris-wood: Yep, missed that!

src/crypto/tls/tls_test.go
chris-wood: I can whip something up. Stand by.

chris-wood:
1. I vote for just logging the event upon completion. We can always add the event in alert code paths later if needed.
2. I left this as read, since that implied (to me) reading and processing, but I can change it if you or others feel strongly!
3. Done!

src/crypto/tls/metrics_test.go
```diff
@@ -0,0 +1,139 @@
...
+import (
+	"crypto/x509"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"testing"
+	"time"
+)
+
+func (e EXP_EventTLS13ServerHandshakeTimingInfo) isMonotonic() bool {
```
cjpatton, line 15: 1. Suggestion: Instead of making this a method of of `EXP_EventTLS13Server...` I would just hard-code the checks in the test below. That way no one in the future stupidly tries to use this test-specific code elsewhere.
1. nit: `isMonotonicallyIncreasing` (here and below)
1. Hrm... this test would pass of all the durations were 0. I wonder if we should check that each duration is *strictly* greater than the last, and not just greater than or equal to the last?

src/crypto/tls/handshake_client.go
```diff
@@ -186,6 +186,9 @@ func (c *Conn) clientHandshake() (err error) {
 		return err
 	}
 
+	handshakeTimings := createTLS13ClientHandshakeTimingInfo(c.config.Time)
```
cjpatton, line 189: `handshakeTimings` should be instantiated at the very beginning of `clientHandshake()`.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -16,6 +16,44 @@ import (
...
+	return "TLS13ClientHandshakeTimingInfo"
+}
+
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) elapsedTime() time.Duration {
+	return time.Now().Sub(e.start)
+}
+
+func createTLS13ClientHandshakeTimingInfo(timerFunc func() time.Time) EXP_EventTLS13ClientHandshakeTimingInfo {
+	timer := time.Now
+	if timerFunc != nil {
```
cjpatton, line 47: `timerFunc` is not equal to `nil` for each call to this function. Since this is an internal API, I would just assume this to always be the case and not do this check here.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -16,6 +16,44 @@ import (
...
+	ReadServerFinished      time.Duration
+	WriteClientFinished     time.Duration
+}
+
+// Name is required by the EXP_Event interface.
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) Name() string {
+	return "TLS13ClientHandshakeTimingInfo"
+}
+
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) elapsedTime() time.Duration {
```
cjpatton, line 41: Just a suggestion: Since this struct only ever exists in the context of a `clientHandshakeStateTLS13`, I would just make `elapsedTime()` a method of `clientHandshakeStateTLS13` so that you can use `hs.c.config.Time()` to get the current time. (Similarly for the server timing info event.)

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -450,6 +494,7 @@ func (hs *clientHandshakeStateTLS13) readServerParameters() error {
 		c.sendAlert(alertUnexpectedMessage)
 		return unexpectedMessageError(encryptedExtensions, msg)
 	}
+
```
cjpatton, line 497: Remove this spurious new line.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -22,6 +22,44 @@ import (
 // messages cause too much work in session ticket decryption attempts.
 const maxClientPSKIdentities = 5
 
+// EXP_EventTLS13ServerHandshakeTimingInfo is carries intra-stack durations for
```
cjpatton, line 25:
```suggestion
// EXP_EventTLS13ServerHandshakeTimingInfo carries intra-stack durations for
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -84,6 +125,7 @@ func (hs *serverHandshakeStateTLS13) handshake() error {
 func (hs *serverHandshakeStateTLS13) processClientHello() error {
 	c := hs.c
 
+	hs.handshakeTimings.ReadClientHello = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 128: This should go after (the HRR and) CH has been processed, i.e., at the end of this function.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -895,6 +947,8 @@ func (hs *serverHandshakeStateTLS13) readClientFinished() error {
 		return unexpectedMessageError(finished, msg)
 	}
 
+	hs.handshakeTimings.ReadClientFinished = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 950: This should come after verifying the MAC (i.e., at the end of this function).

src/crypto/tls/handshake_test.go
```diff
@@ -330,6 +330,25 @@ func allCipherSuites() []uint16 {
 
 var testConfig *Config
 
+// handshakeTestResult captures the test result of a given handshake.
```
cjpatton, line 333: This isn't used anywhere, AFAICT. Delete it?

src/crypto/tls/handshake_test.go
```diff
@@ -393,6 +412,7 @@ func testHandshake(t *testing.T, clientConfig, serverConfig *Config) (serverStat
 	const sentinel = "SENTINEL\n"
 	c, s := localPipe(t)
 	errChan := make(chan error)
+
```
cjpatton, line 415: Remove space.

src/crypto/tls/handshake_test.go
```diff
@@ -412,6 +432,7 @@ func testHandshake(t *testing.T, clientConfig, serverConfig *Config) (serverStat
 		}
 		errChan <- nil
 	}()
+
```
cjpatton, line 435: Remove space.

src/crypto/tls/metrics_test.go
```diff
@@ -0,0 +1,139 @@
...
+	issuer, err := x509.ParseCertificate(testRSACertificateIssuer)
+	if err != nil {
+		panic(err)
+	}
+	rootCAs := x509.NewCertPool()
+	rootCAs.AddCert(issuer)
+
+	const serverName = "example.golang"
+
+	clientConfig := &Config{
```
cjpatton, line 102: Since the `clientConfig` and `serverConfig` are nearly identical, you can start with a `baseConfig` then do `clientConfig := baseConfig.Clone()` and `severConfig := baseConfig.Clone()`. That'll save you some LOC.

src/crypto/tls/metrics_test.go
```diff
@@ -0,0 +1,139 @@
...
+func (e EXP_EventTLS13ClientHandshakeTimingInfo) isMonotonic() bool {
+	return e.WriteClientHello <= e.ReadServerHello &&
+		e.ReadServerHello <= e.ReadEncryptedExtensions &&
+		e.ReadEncryptedExtensions <= e.ReadCertificate &&
+		e.ReadCertificate <= e.ReadCertificateVerify &&
+		e.ReadCertificateVerify <= e.ReadServerFinished &&
+		e.ReadServerFinished <= e.WriteClientFinished
+}
+
+type timingInfo struct {
```
cjpatton, line 33: I would name this something that reflects it's only meant to be used for testing, e.g., `testTimingResult` or something.

src/crypto/tls/metrics_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Fatalf("Handshake failed: %v", err)
+	}
+
+	if !ts.serverTimingInfo.isMonotonic() {
+		t.Fatalf("Server timing information is not monotonic")
+	}
+	if !ts.clientTimingInfo.isMonotonic() {
+		t.Fatalf("Client timing information is not monotonic")
+	}
+}
```
cjpatton, line 139: Do macs not add new lines??

cjpatton (changes requested):
Looking good, just two blocking comments:
1. We forgot about client Certificate and CertificateVerify :/
1. We need to move ReadClientHello until after HRR.

src/crypto/tls/handshake_client_tls13.go
chris-wood: Untrue! Sometimes it is equal to `nil`, depending on what's in the `Config` struct.

src/crypto/tls/handshake_client_tls13.go
chris-wood: This isn't spurious -- I wanted to separate them with whitespace.

src/crypto/tls/handshake_server_tls13.go
chris-wood: 👍

src/crypto/tls/handshake_server_tls13.go
chris-wood: Eh, I'm not sure. There's a lot of stuff in this function that happens as a result of processing the CH. But I'll move it now. We can shuffle it around later.

src/crypto/tls/handshake_server_tls13.go
chris-wood: This one I disagree with. It shouldn't go *quite* at the end.

src/crypto/tls/handshake_test.go
chris-wood: Yep, missed that.

src/crypto/tls/handshake_test.go
chris-wood: I kept it for logical separation.

src/crypto/tls/handshake_test.go
chris-wood: Same as above.

src/crypto/tls/metrics_test.go
chris-wood: I didn't even know `Clone` was a thing. That works.

src/crypto/tls/metrics_test.go
chris-wood: 👍

src/crypto/tls/metrics_test.go
chris-wood: Depends on the editor!

src/crypto/tls/handshake_client_tls13.go
chris-wood: I left this as-is for now, since it seems cleaner for the timing logic to hang off the thing storing the timing, rather than bolted onto something else.

src/crypto/tls/metrics_test.go
chris-wood: I went with a variation of this to keep the test code somewhat leaner.

src/crypto/tls/handshake_client_tls13.go
cjpatton: Our policy has been to avoid these kinds of changes, since they make rebasing harder.

src/crypto/tls/handshake_client_tls13.go
cjpatton: Oh, good catch.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -409,6 +411,7 @@ func (hs *clientHandshakeStateTLS13) processServerHello() error {
 	}
 
 	if !hs.serverHello.selectedIdentityPresent {
+		hs.handshakeTimings.ReadServerHello = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 414: Good catch. To avoid calling this twice, I think we should add the following to the top of `processServerHello()`:
```
defer func() {
    hs.handshakeTimings.ReadServerHello = hs.handshakeTimings.elapsedTime()
}()
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -922,6 +928,8 @@ func (hs *serverHandshakeStateTLS13) readClientCertificate() error {
 		}
 
 		hs.transcript.Write(certVerify.marshal())
+
+		hs.handshakeTimings.ReadCertificateVerify = hs.handshakeTimings.elapsedTime()
```
cjpatton, line 932: I think this should be run unconditionally. (Right now it's only run if there is a client certificate.) Maybe move this  below on line 934?

cjpatton (changes requested):
Just a couple minor things.

src/crypto/tls/handshake_client_tls13.go
chris-wood: I'm not sure. We want this to cover the entire cost of processing the SH, and that means running to the end of each branch.

src/crypto/tls/handshake_server_tls13.go
chris-wood: That's a good idea, yeah. Stand by.

src/crypto/tls/handshake_client_tls13.go
chris-wood: Fixed.

cjpatton (approved):
Looks good!

chris-wood:
@claucece, @Lekensteyn: can you please review?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -16,6 +16,46 @@ import (
 	"time"
 )
 
+// EXP_EventTLS13ClientHandshakeTimingInfo carries intra-stack durations for
+// state machine changes. It can be used for tracking metrics during a
+// connection.
+//
+// NOTE: This API is EXPERIMENTAL and subject to change.
```
claucece, line 23: Maybe the note should also include that it should not be included apart of an experiment as it might leak information @cjpatton

claucece (approved):
LGTM.

src/crypto/tls/handshake_client_tls13.go
chris-wood: I'm not sure anything new is exposed here, but it doesn't hurt to be safe. I added a comment to this effect.

claucece: Thank you!

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -248,6 +293,9 @@ GroupSelection:
 	}
 
 	c.serverName = hs.clientHello.serverName
+
+	hs.handshakeTimings.ReadClientHello = hs.handshakeTimings.elapsedTime()
```
Lekensteyn, line 297: This is too late, it also includes time spend on generating a key share. Maybe ReadClientHello should be called `ProcessClientHello` (or be documented to be more than just receiving the CH)?

src/crypto/tls/handshake_client.go
```diff
@@ -215,15 +219,16 @@ func (c *Conn) clientHandshake() (err error) {
...
+			c:                c,
+			serverHello:      serverHello,
+			hello:            hello,
+			helloInner:       helloInner,
+			helloBase:        helloBase,
+			ecdheParams:      ecdheParams,
+			session:          session,
+			earlySecret:      earlySecret,
+			binderKey:        binderKey,
+			handshakeTimings: handshakeTimings,
```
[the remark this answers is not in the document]
chris-wood: I think it's important to note that this is just for the handshake, so I think ought to keep the name as-is.

src/crypto/tls/handshake_server_tls13.go
chris-wood: I think we want to this in its current location, though of course that may change. Fixed by renaming for now!

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -22,6 +22,48 @@ import (
 // messages cause too much work in session ticket decryption attempts.
 const maxClientPSKIdentities = 5
 
+// EXP_EventTLS13ServerHandshakeTimingInfo carries intra-stack time durations
+// for state machine changes. It can be used for tracking metrics during a
+// connection. Some durations may be sensitive, such as the amount of time to
+// process a particular handshake message, so this event should only be used
+// for experimental purposes.
+//
```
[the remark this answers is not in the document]
chris-wood: I clarified the TLS 1.3-only nature of this API. I also renamed ReadClientHello to ProcessClientHello, so that rolls up the time to deal with HRR.

Lekensteyn (approved):
LGTM

PR #48, opened 2020-12-16
crypto/tls: Fix the testingTriggerHRR condition

cjpatton:
The server sends an HRR if the client does not offer a key share for a key exchange algorithm the server doesn't support. For  testing purposes, it's useful to trigger this codepath manually. If testingTriggerHRR in common.go is set, then the server only advertises support for algorithms that the client supports, but did not provide a key share for.

This change fixes a bug in the trigger logic. It seemed to work for the existing algorithm preferences, but could break if the  preferences change.

cc/ @chris-wood

### DIFF
--- a/src/crypto/tls/handshake_server_tls13.go
+++ b/src/crypto/tls/handshake_server_tls13.go
@@ -223,17 +223,20 @@ func (hs *serverHandshakeStateTLS13) processClientHello() error {
 		// intentionally, we compute the set of ECDHE groups supported by both
 		// the client and server but for which the client did not offer a key
 		// share.
-		supportedCurves = make([]CurveID, 0)
+		m := make(map[CurveID]bool)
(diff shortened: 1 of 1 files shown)


claucece (approved):

claucece:
Should we merge @cjpatton ?

chris-wood:
Good catch. Out of curiosity, what are the groups the client sends in "supported_groups" but does not send a corresponding key share?

chris-wood (approved):

cjpatton:
> Good catch. Out of curiosity, what are the groups the client sends in "supported_groups" but does not send a corresponding key share?

Quite a lot! E.g., P521.

 I'm pretty sure it only sends X25519, though KEM TLS might change this?

chris-wood:
Yeah, it will. I wasn't sure what the supported groups were for Go's implementation. Thanks!

PR #52, opened 2021-01-26
crypto/tls: Fix a bug in client-side ECH logic

cjpatton:
When constructing the ServerHelloConf string, the client may re-order
the ServerHello extensions, causing the client to presume the
ClientHelloOuter was used. Because their transcripts don't match, the
client aborts with a "bad record MAC" alert.

This bug is a result of the manner in which the client computes this
string. It first unmarshals the ServerHello, overwrites the last 8 bytes
of ServerHello.random, then marshals the result the same way the server
marshals its ServerHello, but the extensions may appear in a different
order.

The solution was to write a dedicated function for encoding mapping the
ServerHello to the ServerHelloConf.

### DIFF
--- a/src/crypto/tls/ech.go
+++ b/src/crypto/tls/ech.go
@@ -860,6 +860,62 @@ func echEncodeClientHelloOuterAAD(outerData, handleData []byte) []byte {
 	return outerAadData
 }
 
+// echEncodeServerHelloConf interprets data as a ServerHello message and encodes
+// it as a ServerHelloConf. Returns nil if parsing data fails.
+func echEncodeServerHelloConf(data []byte) []byte {
+	var (
+		msgType                 uint8
+		legacyVersion           uint16
+		random                  []byte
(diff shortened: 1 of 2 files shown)


cjpatton:
(cc/ @Lekensteyn, FYI -- this is a class of bugs I've missed twice now :1st_place_medal:)

src/crypto/tls/ech.go
```diff
@@ -860,6 +860,62 @@ func echEncodeClientHelloOuterAAD(outerData, handleData []byte) []byte {
 	return outerAadData
 }
 
+// echEncodeServerHelloConf interprets data as a ServerHello message and encodes
+// it as a ServerHelloaConf. Returns nil if parsing data fails.
```
chris-wood, line 864:
```suggestion
// it as a ServerHelloConf. Returns nil if parsing data fails.
```

chris-wood (approved):

PR #53, opened 2021-02-20
crypto/tls: Consolidate Cloudflare-specific features

cjpatton:
This PR adds a parameter to `crypto/tls.Config`, called `CFControl`, which will be used to implement Cloudflare-internal functionalities.

This is the second such feature that we've added to the `crypto/tls` package. In particular, the `EXP_EventHandler` callback in `crypto/tls.Config` is used for computing metrics that WE are interested in, but are not likely to be useful for upstream Go. To emphasize this, this PR renames the callback to `CFEventHandler` and moves all implementations to `tls_cf.go`. The implementations are also renamed accordingly.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -742,12 +742,12 @@ type Config struct {
 	// payload.
 	ServerECHProvider ECHProvider
 
-	// EXP_EventHandler, if set, is called by the client and server at various
+	// CFEventHandler, if set, is called by the client and server at various
 	// points during the handshake to handle specific events. For example, this
 	// callback can be used to record metrics.
 	//
 	// NOTE: This API is EXPERIMENTAL and subject to change.
-	EXP_EventHandler func(event EXP_Event)
+	CFEventHandler func(event CFEvent)
 
 	// mutex protects sessionTicketKeys and autoSessionTicketKeys.
 	mutex sync.RWMutex
@@ -838,7 +838,7 @@ func (c *Config) Clone() *Config {
 		ECHEnabled:                  c.ECHEnabled,
 		ClientECHConfigs:            c.ClientECHConfigs,
 		ServerECHProvider:           c.ServerECHProvider,
-		EXP_EventHandler:            c.EXP_EventHandler,
+		CFEventHandler:              c.CFEventHandler,
 		sessionTicketKeys:           c.sessionTicketKeys,
 		autoSessionTicketKeys:       c.autoSessionTicketKeys,
 	}
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -328,6 +328,13 @@ type ConnectionState struct {
 	// accepted by the server.
 	ECHAccepted bool
 
+	// CFControl is used to pass additional TLS configuration information to
+	// HTTP requests.
+	//
+	// NOTE: This feature is used to implement Cloudflare-internal features.
+	// This feature is unstable and applications MUST NOT depend on it.
+	CFControl interface{}
+
 	// ekm is a closure exposed via ExportKeyingMaterial.
 	ekm func(label string, context []byte, length int) ([]byte, error)
 }
@@ -743,9 +750,19 @@ type Config struct {
 	ServerECHProvider ECHProvider
 
 	// CFEventHandler, if set, is called by the client and server at various
-	// points during the handshake to handle specific events. For example, this
-	// callback can be used to record metrics.
+	// points during the handshake to handle specific events. This is used
+	// primarily for collecting metrics.
+	//
+	// NOTE: This feature is used to implement Cloudflare-internal features.
+	// This feature is unstable and applications MUST NOT depend on it.
 	CFEventHandler func(event CFEvent)
+
+	// CFControl is used to pass additional TLS configuration information to
+	// HTTP requests via ConnectionState.
+	//
+	// NOTE: This feature is used to implement Cloudflare-internal features.
+	// This feature is unstable and applications MUST NOT depend on it.
+	CFControl interface{}
 
 	// mutex protects sessionTicketKeys and autoSessionTicketKeys.
 	mutex sync.RWMutex
@@ -837,6 +854,7 @@ func (c *Config) Clone() *Config {
 		ClientECHConfigs:            c.ClientECHConfigs,
 		ServerECHProvider:           c.ServerECHProvider,
 		CFEventHandler:              c.CFEventHandler,
+		CFControl:                   c.CFControl,
 		sessionTicketKeys:           c.sessionTicketKeys,
 		autoSessionTicketKeys:       c.autoSessionTicketKeys,
 	}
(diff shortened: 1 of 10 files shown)


src/crypto/tls/tls_cf.go
```diff
@@ -48,8 +56,144 @@ func init() {
...
+// connection. Some durations may be sensitive, such as the amount of time to
+// process a particular handshake message, so this event should only be used
+// for experimental purposes.
+type CFEventTLS13ClientHandshakeTimingInfo struct {
+	timer                   func() time.Time
+	start                   time.Time
+	WriteClientHello        time.Duration
+	ProcessServerHello      time.Duration
+	ReadEncryptedExtensions time.Duration
+	ReadCertificate         time.Duration
```
claucece, line 77: Is `ReadCertificate` both used for client and server?

src/crypto/tls/tls_cf.go
```diff
@@ -1,3 +1,6 @@
+// Copyright 2021 Cloudflare, Inc. All rights reserved. Use of this source code
+// is governed by a BSD-style license that can be found in the LICENSE file.
+
```
claucece, line 3: Which remains me that we should also update the other files that have this copyright to '2020-2021'

claucece:
LGTM. Just added some questions.

src/crypto/tls/tls_cf.go
cjpatton: There are two events, one that is emitted by the client and one that is emitted by the server. Both record this value.

src/crypto/tls/tls_cf.go
cjpatton: TBH I have no idea what I'm doing with the copyright messages! I know how BSD license works when it's my own encode, but since we're extending someone else's work (i.e., the Go team's), I'm not sure if we get to hold copyright on this file or not.

claucece: We should really define this ;) Maybe we can have a short call for the specific bureaucracy issues of cf-go?

src/crypto/tls/common.go
```diff
@@ -328,6 +328,14 @@ type ConnectionState struct {
 	// accepted by the server.
 	ECHAccepted bool
 
+	// CFFlags is used to pass additional TLS configuration information to HTTP
+	// requests.
+	//
+	// NOTE: This feature is used to implement Cloudflare-internal features.
+	// Applications should consider this feature unstable and avoid depending on
+	// it.
```
chris-wood, line 336:
```suggestion
	// Applications MUST NOT depend on the semantics of this field.
```

chris-wood: (Here and below)

src/crypto/tls/tls_cf.go
chris-wood: Yeah, though we can address this in a followup change.

src/crypto/tls/tls_cf.go
```diff
@@ -46,10 +57,146 @@ func init() {
...
-// NOTE: This API is EXPERIMENTAL and subject to change.
-type EXP_Event interface {
+// CFEvent is a value emitted at various points in the handshake that is
+// handled by the callback Config.CFEventHandler.
+type CFEvent interface {
 	Name() string
 }
+
+// CFEventTLS13ClientHandshakeTimingInfo carries intra-stack time durations
+// for TLS 1.3 state machine changes. It can be used for tracking metrics during a
```
chris-wood, line 67:
```suggestion
// for TLS 1.3 client state machine changes. It can be used for tracking metrics during a
```

src/crypto/tls/tls_cf.go
```diff
@@ -46,10 +57,146 @@ func init() {
...
+	}
+
+	return CFEventTLS13ClientHandshakeTimingInfo{
+		timer: timer,
+		start: timer(),
+	}
+}
+
+// CFEventTLS13ServerHandshakeTimingInfo carries intra-stack time durations
+// for TLS 1.3 state machine changes. It can be used for tracking metrics during a
```
chris-wood, line 107:
```suggestion
// for TLS 1.3 server state machine changes. It can be used for tracking metrics during a
```

src/crypto/tls/tls_cf_test.go
```diff
@@ -0,0 +1,14 @@
+package tls
+
+import (
+	"testing"
+)
+
+// Check that CFFlags is correctly propagated from Config to ConnectionState.
+func TestPropagateCFFlags(t *testing.T) {
+	want := uint64(23)
```
chris-wood, line 9: Not 42?!

chris-wood (approved):
Nice consolidation and refactor 👍

Lekensteyn:
The changes look fine except for one thing: the new CFFlags uint64 field.

So the intended purpose appears to be passing TLS connection level information to the HTTP handler. More specifically, state from the GetCertificate callback is propagated.

The current choice of the uint64 type is very limiting, what if someone would like to make the Server Certificate available in the HTTP handler, for example to enforce that the Host header is covered by any of the Certificate SANs for connection coalescing? (hostcheck.go in tls-1000)

I would recommend using contexts to pass stuff from TLS -> HTTP handler, something like https://github.com/cloudflare/go/commit/88253a956a753213617d95af3f42a23a78798473. That commit was used to expose 0-RTT state (which was available on the `tls.Conn` struct), but we could also expose `tls.Config` or other types.

Lekensteyn (changes requested):
Drop CFFlags, it might not be necessary.

chris-wood:
> Drop CFFlags, it might not be necessary.

Moving to a context table seems like something we could change on top of this refactor, right? (That change would only affect code that touches the flags, and would be quite small in comparison.) So if there are no other objections, I'd say let's merge this and then build on top.

src/crypto/tls/tls_cf.go
cjpatton: I'll bring it up in the t1000 meeting today.

src/crypto/tls/tls_cf_test.go
cjpatton: I am *not* thankful for any fish.

cjpatton:
> The changes look fine except for one thing: the new CFFlags uint64 field.
> 
> So the intended purpose appears to be passing TLS connection level information to the HTTP handler. More specifically, state from the GetCertificate callback is propagated.

Not exclusively. We also need to propagate zone-level configuration beyond what GetCertificate() provides.

> The current choice of the uint64 type is very limiting, what if someone would like to make the Server Certificate available in the HTTP handler, for example to enforce that the Host header is covered by any of the Certificate SANs for connection coalescing? (hostcheck.go in tls-1000)
> 
> I would recommend using contexts to pass stuff from TLS -> HTTP handler, something like [88253a9](https://github.com/cloudflare/go/commit/88253a956a753213617d95af3f42a23a78798473). That commit was used to expose 0-RTT state (which was available on the `tls.Conn` struct), but we could also expose `tls.Config` or other types.

I agree that this is limiting, and I was hoping someone would suggest something more general :) I didn't know context. could be used this way. I'm looking into how this would work. The linked commit doesn't do what exactly what we need: it's not sufficient to propagate the `crypto/tls.Conn`.

cjpatton:
I changed `CFFlags uint64` to `CFControl interface{}`.

Lekensteyn (approved):

Issue #62, opened 2021-03-10
Verify tests for DC


armfazh:
Tests for DCs fail when running isolated, but they pass when invoked as a batch (i.e. `go test -v -run=.`)

### What did you do?
```sh
$ cd src/crypto/tls
$ ../../../bin/go test -v -run=DC
```

### What did you expect to see?

```
=== RUN   TestDCHandshakeServerAuth
--- PASS: TestDCHandshakeServerAuth (0.14s)
=== RUN   TestDCHandshakeClientAuth
--- PASS: TestDCHandshakeClientAuth (0.14s)
=== RUN   TestDCHandshakeClientAndServerAuth
--- PASS: TestDCHandshakeClientAndServerAuth (0.03s)

```

### What did you see instead?
```
    delegated_credentials_test.go:592: test #1 dcCounter:1, (tls13: DC server and client support) fails: x509: certificate signed by unknown authority
    delegated_credentials_test.go:598: test #1 (tls13: DC server and client support) usedDC = false; expected true
    delegated_credentials_test.go:592: test #1 dcCounter:2, (tls13: DC server and client support) fails: x509: certificate signed by unknown authority
    delegated_credentials_test.go:598: test #1 (tls13: DC server and client support) usedDC = false; expected true
    delegated_credentials_test.go:592: test #1 dcCounter:3, (tls13: DC server and client support) fails: x509: certificate signed by unknown authority
    delegated_credentials_test.go:598: test #1 (tls13: DC server and client support) usedDC = false; expected true

```

claucece:
Yes, I see. I'm not really sure why is only flagged in this case of testing but I can take a look at this next week. 

The problem is that we are providing a mock of a certificate (without the correct root and chain and all), as it is only for testing.

We can provide that whole chain, though it might be too much for testing. Or just set the `c.config.InsecureSkipVerify` to true, as it is often on testing, which will invalidate the check (as it fails on `verifyServerCertificate`).

claucece:
I added a PR @armfazh ;) Let me know your thoughts.

claucece:
Merged now.

PR #64, opened 2021-03-15
Add insecure verify #62

claucece:
Adding to allow usage of mock certs for testing.

cc./ @armfazh

### DIFF
--- a/src/crypto/tls/delegated_credentials_test.go
+++ b/src/crypto/tls/delegated_credentials_test.go
@@ -570,6 +570,7 @@ func TestDCHandshakeServerAuth(t *testing.T) {
 	clientConfig := dcTestConfig.Clone()
 	serverConfig := dcTestConfig.Clone()
 	serverConfig.GetCertificate = testServerGetCertificate
+	clientConfig.InsecureSkipVerify = true
 
 	for i, test := range dcTests {
 		clientConfig.SupportDelegatedCredential = test.clientDCSupport


armfazh:
what is the effect in DCs when this flags is enabled?
in the tests, we want that the verification of DCs be tested too.

claucece:
> what is the effect in DCs when this flags is enabled?

Verifies the root and the chain. So you will have to provide the cert chain.

> in the tests, we want that the verification of DCs be tested too.

why would you want it if you are testing with a mock certificate?


EDIT: DC are only placed on the leaf (end-entity) certificate: this means that only the leaf and the DC have to be valid for testing purposes. The verification of the whole certificate chain is already tested in other parts of crypto/tls.

armfazh:
> EDIT: DC are only placed on the leaf (end-entity) certificate: this means that only the leaf and the DC have to be valid for testing purposes. The verification of the whole certificate chain is already tested in other parts of crypto/tls.

Thanks for clarifying this to me.

armfazh (approved):

Lekensteyn:
> why would you want it if you are testing with a mock certificate?

It is more realistic, perhaps the implementation treats InsecureSkipVerify specially and skips DC validation too. However given that your tests already seem to cover negative cases too for DC, it probably doesn't matter. Thanks for the clarification!

Lekensteyn (approved):

claucece:
> It is more realistic, perhaps the implementation treats InsecureSkipVerify specially and skips DC validation too. However given that your tests already seem to cover negative cases too for DC, it probably doesn't matter. Thanks for the clarification!

Yes! We have some lengthy debate around InsecureSkipVerify skipping the verification for a DC too: as it is even with InsecureSkipVerify set to true, the DC will always be validated.

Issue #67, opened 2021-03-15
Define an API for DCs


claucece:
Currently, there is a callback for using DC: `GetDelegatedCredential`. This can be improved to:

- [x] Use the `GetCertificate` and `GetCertificateClient` callbacks to return the DCs as part of the Certificate structure
- [x] Allow the possibility of returning a list of DCs: the appropriate DC will be chosen by using the advertise signature algorithm on the ClientHello/Certificate Request extension.
- [x] Refactor the DC tests.

claucece:
cc./ @armfazh @wbl

claucece:
Related to #63

cjpatton:
I like the idea of adding DCs to the `crypto/tls.Certificate` structure, since it allows us to expose DCs by whatever means is easiest to configure. Something like this?
```go
type Certificate struct {
   ...
   DCs [][]byte
   DCPrivateKeys []crypto.PrivateKey
}
```
so that `DCPrivateKeys[i]` is the private key of `DCs[i]` for all `0 <= i <= len(DCs)`?

claucece:
@cjpatton check the latest pr: https://github.com/cloudflare/go/commit/dc791fdbeeacff13d3e1f076106a013fd56cd5f3

cjpatton:
You're idea is much better than mine! I left some comments on the PR.

claucece:
Merged now. Closing this.

Issue #63, opened 2021-03-10
exhaustive checks on DC tests


armfazh:
In TestDCHandshakeServerAuth, the external loop iterates over the dcTests, and the internal loops over the signatureSchemes. It seems that the test is not executing all the combinations of (dcTests, signatureSchemes), is this the intended test?
Inspecting the dcCounter value, it runs from 0 to Len only for the first test case.

```go
	for i, test := range dcTests {
		clientConfig.SupportDelegatedCredential = test.clientDCSupport
+		t.Logf("dcCounter: %v\n", dcCounter)
		for dcCounter < len(dcTestDCScheme)-1 {
			if test.serverDCSupport {
				serverConfig.GetDelegatedCredential = testGetDelegatedCredential
				dcCounter++
			} else {
				serverConfig.GetDelegatedCredential = nil
			}
```

claucece:
Yes, it is expected. It is really not needed to test all the conditions as the other ones will fail one way or another. The added of a `dcCounter` was a request, which I think is unnecessary as it tests with every possible algorithm the tls connection (while in truthness if the DC is able to be created and validated, then it should work).

I expected to change the way DC test work when the new API lands (next week when I return to work): the usage of callbacks make it almost impossible to check with different algos and variables.

claucece:
I'm going to leave this open for a while until the design of the new API is taken into account.

armfazh:
> the usage of callbacks make it almost impossible to check with different algos and variables

it is possible to create a factory of functions, so each produced function must use one combination of dc and signature scheme.

claucece:
> it is possible to create a factory of functions, so each produced function must use one combination of dc and signature scheme.

The callbacks will likely disappear in the new API by using the cert ones, so no need of it then.

claucece:
Solved now.

PR #69, opened 2021-03-18
Define api for delegated credentials #67

claucece:
Ready for review.

This should solve #63 as well.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -497,7 +497,8 @@ type ClientHelloInfo struct {
 
 	// SignatureSchemesDC lists the signature schemes that the client
 	// is willing to verify when using Delegated Credentials.
-	// This can be different from SignatureSchemes.
+	// This is and can be different from SignatureSchemes. SignatureSchemesDC
+	// is set only if the DelegatedCredentials Extension is being used.
 	// If Delegated Credentials are supported, this list should not be nil.
 	SignatureSchemesDC []SignatureScheme
 
@@ -549,6 +550,9 @@ type CertificateRequestInfo struct {
 
 	// SignatureSchemesDC lists the signature schemes that the server
 	// is willing to verify when using Delegated Credentials.
+	// This is and can be different from SignatureSchemes. SignatureSchemesDC
+	// is set only if the DelegatedCredentials Extension is being used.
+	// If Delegated Credentials are supported, this list should not be nil.
 	SignatureSchemesDC []SignatureScheme
 
 	// Version is the TLS version that was negotiated for this connection.
@@ -820,13 +824,6 @@ type Config struct {
 	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
 	SupportDelegatedCredential bool
 
-	// GetDelegatedCredential returns a DelegatedCredential for use with the
-	// delegated credential extension based on the ClientHello. It only works
-	// with TLS 1.3. If this is nil, then the server will not offer
-	// a DelegatedCredential. If the call returns nil, the server is also
-	// not offering a DelegatedCredential.
-	GetDelegatedCredential func(*ClientHelloInfo, *CertificateRequestInfo) (*DelegatedCredential, crypto.PrivateKey, error)
-
 	// mutex protects sessionTicketKeys and autoSessionTicketKeys.
 	mutex sync.RWMutex
 	// sessionTicketKeys contains zero or more ticket keys. If set, it means the
@@ -917,7 +914,6 @@ func (c *Config) Clone() *Config {
 		Renegotiation:               c.Renegotiation,
 		KeyLogWriter:                c.KeyLogWriter,
 		SupportDelegatedCredential:  c.SupportDelegatedCredential,
-		GetDelegatedCredential:      c.GetDelegatedCredential,
 		ECHEnabled:                  c.ECHEnabled,
 		ClientECHConfigs:            c.ClientECHConfigs,
 		ServerECHProvider:           c.ServerECHProvider,
@@ -1462,6 +1458,16 @@ func (c *Config) writeKeyLog(label string, clientRandom, secret []byte) error {
 // and is only for debugging, so a global mutex saves space.
 var writerMutex sync.Mutex
 
+// A DelegatedCredentialPair contains a Delegated Credential and its
+// associated private key.
+type DelegatedCredentialPair struct {
+	// DC is the delegated credential.
+	DC *DelegatedCredential
+	// PrivateKey is the private key used to derive the public key of
+	// contained in DC. PrivateKey must implement crypto.Signer.
+	PrivateKey crypto.PrivateKey
+}
+
 // A Certificate is a chain of one or more certificates, leaf first.
 type Certificate struct {
 	Certificate [][]byte
@@ -1479,9 +1485,15 @@ type Certificate struct {
 	// SignedCertificateTimestamps contains an optional list of Signed
 	// Certificate Timestamps which will be served to clients that request it.
 	SignedCertificateTimestamps [][]byte
-	// DelegatedCredential is a serialized Delegated Credential, signed by
-	// the leaf certificate.
-	// If not supported, this will be nil.
+	// DelegatedCredentials are a list of Delegated Credentials with their
+	// corresponding private keys, signed by the leaf certificate.
+	// If there are no delegated credentials, this field is nil.
+	DelegatedCredentials []DelegatedCredentialPair
+	// DelegatedCredential is the delegated credential to be used in the
+	// handshake.
+	// If there are no delegated credentials, this field is nil.
+	// NOTE: Do not fill this field, as it will be filled depending on
+	// the provided list of delegated credentials.
 	DelegatedCredential []byte
 	// Leaf is the parsed form of the leaf certificate, which may be initialized
 	// using x509.ParseCertificate to reduce per-handshake processing. If nil,
--- a/src/crypto/tls/delegated_credentials_test.go
+++ b/src/crypto/tls/delegated_credentials_test.go
@@ -17,13 +17,6 @@ import (
 	"time"
 )
 
-// dcAndPrivateKey stores a Delegated Credential and its corresponding private
-// key.
-type dcAndPrivateKey struct {
-	*DelegatedCredential
-	privateKey crypto.PrivateKey
-}
-
 // These test keys were generated with the following program, available in the
 // crypto/tls directory:
 //
@@ -137,12 +130,12 @@ ZyJKvc2KqjGeZh0Or5pq6ZJb0zR7WPdz5aJIzaZ5YcxLMSv0KwaAEPH2
 `
 
 var (
-	dcTestConfig   *Config
-	dcTestCerts    map[string]*Certificate
-	serverDC       []*dcAndPrivateKey
-	clientDC       []*dcAndPrivateKey
-	dcNow          time.Time
-	dcTestDCScheme = []SignatureScheme{ECDSAWithP256AndSHA256, ECDSAWithP384AndSHA384, ECDSAWithP521AndSHA512, Ed25519}
+	dcTestConfig            *Config
+	dcTestCerts             map[string]*Certificate
+	serverDC                []DelegatedCredentialPair
+	clientDC                []DelegatedCredentialPair
+	dcNow                   time.Time
+	dcTestDCSignatureScheme = []SignatureScheme{ECDSAWithP256AndSHA256, Ed25519, ECDSAWithP384AndSHA384, ECDSAWithP521AndSHA512}
 )
 
 func init() {
@@ -232,7 +225,7 @@ func initDCTest() {
 	}
 	dcTestCerts["no dc"] = noDcCert
 
-	// The root certificates for the client.
+	// The root certificates for the peer.
 	dcTestConfig.RootCAs = x509.NewCertPool()
 
 	for _, c := range dcTestCerts {
@@ -243,20 +236,19 @@ func initDCTest() {
 		dcTestConfig.RootCAs.AddCert(dcRoot)
 	}
 
-	for i := 0; i < len(dcTestDCScheme); i++ {
-		dc, sk, err := NewDelegatedCredential(dcCertP256, dcTestDCScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, false)
+	for i := 0; i < len(dcTestDCSignatureScheme); i++ {
+		dc, priv, err := NewDelegatedCredential(dcCertP256, dcTestDCSignatureScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, false)
 		if err != nil {
 			panic(err)
 		}
-		serverDC = append(serverDC, &dcAndPrivateKey{dc, sk})
+		serverDC = append(serverDC, DelegatedCredentialPair{dc, priv})
 
-		dc, sk, err = NewDelegatedCredential(dcCertP256, dcTestDCScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, true)
+		dc, priv, err = NewDelegatedCredential(dcCertP256, dcTestDCSignatureScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, true)
 		if err != nil {
 			panic(err)
 		}
-		clientDC = append(clientDC, &dcAndPrivateKey{dc, sk})
+		clientDC = append(clientDC, DelegatedCredentialPair{dc, priv})
 	}
-
 }
 
 func publicKeysEqual(publicKey, publicKey2 crypto.PublicKey, algo SignatureScheme) error {
@@ -387,7 +379,7 @@ func TestDelegatedCredentialMarshal(t *testing.T) {
 	cert := dcTestCerts["dcEd25519"]
 	time := dcNow.Sub(cert.Leaf.NotBefore) + dcMaxTTL
 
-	for _, sig := range dcTestDCScheme {
+	for _, sig := range dcTestDCSignatureScheme {
 		delegatedCred, _, err := NewDelegatedCredential(cert, sig, time, false)
 		if err != nil {
 			t.Fatal(err)
@@ -418,8 +410,23 @@ func TestDelegatedCredentialMarshal(t *testing.T) {
 	}
 }
 
-var dcTests = []struct {
+var dcServerTests = []struct {
 	clientDCSupport bool
+	clientMaxVers   uint16
+	serverMaxVers   uint16
+	expectSuccess   bool
+	expectDC        bool
+	name            string
+}{
+	{true, VersionTLS13, VersionTLS13, true, true, "tls13: DC client support"},
+	{false, VersionTLS13, VersionTLS13, true, false, "DC not client support"},
+	{true, VersionTLS12, VersionTLS13, true, false, "client using TLS 1.2. No DC is supported in that version."},
+	{true, VersionTLS13, VersionTLS12, true, false, "server using TLS 1.2. No DC is supported in that version."},
+	{true, VersionTLS11, VersionTLS13, true, false, "client using TLS 1.1. No DC is supported in that version."},
+	{true, VersionTLS13, VersionTLS10, false, false, "server using TLS 1.0. No DC is supported in that version."},
+}
+
+var dcClientTests = []struct {
 	serverDCSupport bool
 	clientMaxVers   uint16
 	serverMaxVers   uint16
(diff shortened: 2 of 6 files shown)


claucece:
@armfazh @lukevalenta @Lekensteyn ready for review now. Thanks!

src/crypto/tls/common.go
```diff
@@ -1462,6 +1458,16 @@ func (c *Config) writeKeyLog(label string, clientRandom, secret []byte) error {
 // and is only for debugging, so a global mutex saves space.
 var writerMutex sync.Mutex
 
+// A DelegatedCredentialStruct contains a Delegated Credential and its
+// associated private key.
+type DelegatedCredentialStruct struct {
```
cjpatton: I would call this `DelegatedCredentialKeyPair`, or even `DCKeyPair`.

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -428,9 +420,27 @@ var dcTests = []struct {
 	name            string
 }{
 	{true, true, VersionTLS13, VersionTLS13, true, true, "tls13: DC server and client support"},
-	{true, false, VersionTLS13, VersionTLS13, true, false, "DC not server support"},
+	{true, false, VersionTLS13, VersionTLS13, true, true, "DC not server support"},
 	{false, true, VersionTLS13, VersionTLS13, true, false, "DC not client support"},
 	{true, true, VersionTLS12, VersionTLS13, true, false, "client using TLS 1.2. No DC is supported in that version."},
+	{true, true, VersionTLS13, VersionTLS12, true, false, "server using TLS 1.2. No DC is supported in that version."}, // TODO: this
```
cjpatton: > // TODO: this

Is this TODO meant to be resolved before merging? If not, please self-assign it (e.g., TODO(claucece))

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -562,45 +546,39 @@ func testConnWithDC(t *testing.T, clientMsg, serverMsg string, clientConfig, ser
 	return false, nil
 }
 
+var count int
```
cjpatton: This global counter makes me a bit nervous. Does it need to be global, or could it live in some function's local scope? If it must be global, could you add a comment documenting what it does, and which functions write to and read from it?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -709,6 +709,30 @@ func certificateRequestInfo(certReq *certificateRequestMsgTLS13, vers uint16) *C
...
+	if len(cert.DelegatedCredentials) == 0 {
+		return nil, errors.New("No Delegated Credential found.")
+	}
+
+	if len(cert.DelegatedCredentials) == 1 {
+		// There's only one choice, so no point doing any work.
+		return &cert.DelegatedCredentials[0], nil
+	}
+
+	for _, dcStruct := range cert.DelegatedCredentials {
```
cjpatton: `dcKeyPair`?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -723,26 +747,25 @@ func (hs *clientHandshakeStateTLS13) sendClientCertificate() error {
...
 	}
 
-	var dc *DelegatedCredential
-	if hs.certReq.supportDelegatedCredential && c.config.GetDelegatedCredential != nil {
-		var priv crypto.PrivateKey
+	var dcStruct *DelegatedCredentialStruct
+	if hs.certReq.supportDelegatedCredential && len(hs.certReq.supportedSignatureAlgorithmsDC) > 0 {
 		var err error
-		dc, priv, err = c.config.GetDelegatedCredential(nil, cri)
+		delegatedCredentialStruct, err := getClientDelegatedCredential(cri, cert)
```
cjpatton: `dcKeyPair`?

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -445,6 +445,32 @@ func cloneHash(in hash.Hash, h crypto.Hash) hash.Hash {
 	return out
 }
 
+// getDelegatedCredential returns the best Delegated Credential for the given
+// ClientHelloInfo, defaulting to the first element of cert.DelegatedCredentialStruct.
+func getDelegatedCredential(clientHello *ClientHelloInfo, cert *Certificate) (*DelegatedCredentialStruct, error) {
```
cjpatton: Perhaps rename this to `getServerDelegatedCredential` in order to align it with `getClientDelegatedCredential`.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -709,6 +709,30 @@ func certificateRequestInfo(certReq *certificateRequestMsgTLS13, vers uint16) *C
...
 
+func getClientDelegatedCredential(cri *CertificateRequestInfo, cert *Certificate) (*DelegatedCredentialStruct, error) {
+	if len(cert.DelegatedCredentials) == 0 {
+		return nil, errors.New("No Delegated Credential found.")
+	}
+
+	if len(cert.DelegatedCredentials) == 1 {
+		// There's only one choice, so no point doing any work.
+		return &cert.DelegatedCredentials[0], nil
+	}
```
cjpatton, line 726: What happens if the client uses a DC whose algorithm isn't supported by the server?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -709,6 +709,30 @@ func certificateRequestInfo(certReq *certificateRequestMsgTLS13, vers uint16) *C
...
+		// schemes we can use with this delegated credential.
+		if len(cri.SignatureSchemesDC) > 0 {
+			if _, err := selectSignatureSchemeDC(VersionTLS13, dcStruct.DC, cri.SignatureSchemesDC); err == nil {
+				return &dcStruct, nil
+			}
+		}
+	}
+
+	// If nothing matches, return the first delegated credential.
+	return &cert.DelegatedCredentials[0], nil
```
cjpatton: AFAICT, the client may use a DC whose algorithm isn't supported by the server. It seems to me that our options are either to fallback to signing with the leaf certificate or abort with "handshake_failure" if we don't have the signing key for the leaf certificate.

Similar comments for `getDelegatedCredential` below.

cjpatton (commented):
I like the API changes. One concern sticks out to me: If I'm understanding the code correctly, the client (resp. server) may use a DC whose signature algorithm isn't supported by the peer. I may be wrong :)

src/crypto/tls/common.go
claucece: Maybe `DelegatedCredentialPair`? Because it is not the keypair of the delegated credential, but rather the DC and the priv key...

src/crypto/tls/delegated_credentials_test.go
claucece: It was supposed to not be there ahaha ;)

src/crypto/tls/delegated_credentials_test.go
claucece: It needs to be global to be an input to the callbacks for the certificate. The reason why it is is because the `getCertificate` callback does no accept any input other than the clienthelloinfo or criinfo. Let me see if I can change the design...

src/crypto/tls/handshake_server_tls13.go
claucece: Yeah, I was thinking on that too. But the calls for certificate are: `getCertificate` and `getClientCertificate`, so I'm following the naming convention of the library. I think this naming occurs due to the `getCertificate` being the default one and if you really want client auth then it is explicitly defined.

src/crypto/tls/handshake_client_tls13.go
claucece: Good question! So, these two functions `getDelegatedCredential` and `getClientDelegatedCredential` emulate `getCertificate` and `getClientCertificate`: their aim is to return one credential/certificate even if it is not "valid" (I think the logic in Golang was that these functions should just return a certificate).  In the actual handshake, the check will be done by selecting the signature algorithm: https://github.com/cloudflare/go/blob/dc-api/src/crypto/tls/handshake_server_tls13.go#L522 If it is unsupported, it will error.

src/crypto/tls/handshake_client_tls13.go
claucece: This function will return a delegated credential, but at the handshake level, the check will be performed: https://github.com/cloudflare/go/blob/dc-api/src/crypto/tls/handshake_server_tls13.go#L522 https://github.com/cloudflare/go/blob/dc-api/src/crypto/tls/handshake_client_tls13.go#L806 

It errors because the usage of a delegated credential was advertised, a dc was found but it is invalid. It could be that we fallback to the certificate in that case as well, as the draft says:

```
If the extension is present, the server MAY send a delegated
credential; if the extension is not present, the server MUST NOT send
a delegated credential.
```

so it is always a MAY.

src/crypto/tls/handshake_client_tls13.go
claucece: Maybe `dcPair`?

src/crypto/tls/common.go
cjpatton: Works for me! It's just that "Struct" is kind of vague. "Pair" works great.

src/crypto/tls/delegated_credentials_test.go
cjpatton: I see. I would be fine with adding a comment documenting its purpose. Also, it would probably be useful for posterity if the name were more specific than "count".

src/crypto/tls/handshake_client_tls13.go
cjpatton: SGTM

src/crypto/tls/handshake_server_tls13.go
cjpatton: I agree, better to align with the existing terminology.

src/crypto/tls/handshake_client_tls13.go
cjpatton: Ah, right you are. IMO it would be cleaner to return `(nil, nil)` if no compatible DC is found. This should have the same affect, since the branch here wouldn't be taken: https://github.com/cloudflare/go/blob/dc-api/src/crypto/tls/handshake_server_tls13.go#L514

I think this is better because the caller is guaranteed that *if* a DC is returned, then it can be used in the handshake. Alternatively, I would suggest adding a comment to `get[Client]DelegatedCredential` noting that the returned DC may not be suitable for the handshake.

src/crypto/tls/handshake_client_tls13.go
cjpatton: Thanks for clarifying.

src/crypto/tls/handshake_client_tls13.go
claucece: Addressed now.

src/crypto/tls/delegated_credentials_test.go
claucece: I added the comment and changed the var name ;)

claucece:
@cjpatton addressed now. Let me know what you think now ;)

> One concern sticks out to me: If I'm understanding the code correctly, the client (resp. server) may use a DC whose signature algorithm isn't supported by the peer. I may be wrong :)

When you get a DC from storage you check which one you will used based on the other's peer signature algorithm advertised by either the clientHelloInfo or certificateRequestInfo, so it will not be able to use anything that wasn't advertised there in the handshake.

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -428,14 +420,38 @@ var dcTests = []struct {
...
+	{true, true, VersionTLS13, VersionTLS13, true, true, "tls13: DC server and client support"},
+	{true, false, VersionTLS13, VersionTLS13, true, false, "DC not server support"},
+	{false, true, VersionTLS13, VersionTLS13, true, true, "DC not client support"},
+	{true, true, VersionTLS12, VersionTLS13, true, false, "client using TLS 1.2. No DC is supported in that version."},
+	{true, true, VersionTLS13, VersionTLS12, true, false, "server using TLS 1.2. No DC is supported in that version."},
+	{true, true, VersionTLS11, VersionTLS13, true, false, "client using TLS 1.1. No DC is supported in that version."},
+	{true, true, VersionTLS13, VersionTLS10, false, false, "server using TLS 1.0. No DC is supported in that version."},
+}
+
+// dcCount defines the delegated credenatial to be used as returned by the
```
Lekensteyn:
```suggestion
// dcCount defines the delegated credential to be used as returned by the
```

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -446,7 +462,9 @@ func testServerGetCertificate(ch *ClientHelloInfo) (*Certificate, error) {
 	}
 
 	if versOk && ch.SupportsDelegatedCredential {
-		return dcTestCerts["dcP256"], nil
+		serverCert := dcTestCerts["dcP256"]
+		serverCert.DelegatedCredentials = serverDC[dcCount:]
```
Lekensteyn, line 462: (Note for my understanding: `initDCTest` resets `dcTestCerts` and `dcTestCerts["dcCertP256"]` which is why it is okay for `testServerGetCertificate` to have side-effects such as overwriting `dcTestCerts["dcP256"].DelegatedCredentials`.)

I was wondering why `serverDC[dcCount:]` was used. Would `serverDC[dcCount:dcCount+1]` be sufficient to limit the test to a specific DC? Or is there a reason to include more than one?

Lekensteyn (commented):
If I understand correctly, the point of this change is to provide the DC in a single callback via GetCertificate rather than two (GetCertificate followed by GetDelegatedCredential)?

Could you include the motivation for the change in the commit message? Thanks, it looks good with minor comments!

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -428,14 +420,38 @@ var dcTests = []struct {
 	name            string
 }{
 	{true, true, VersionTLS13, VersionTLS13, true, true, "tls13: DC server and client support"},
-	{true, false, VersionTLS13, VersionTLS13, true, false, "DC not server support"},
+	{true, false, VersionTLS13, VersionTLS13, true, true, "DC not server support"},
```
cjpatton: The client indicates support for DCs, but the server does not. And yet a DC was verified by the client? If I'm reading this correctly, I don't think this is the behavior we should expect. In particular, this `true` should be a `false` AFAICT.

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -428,14 +420,38 @@ var dcTests = []struct {
...
+	serverDCSupport bool
+	clientMaxVers   uint16
+	serverMaxVers   uint16
+	expectSuccess   bool
+	expectDC        bool
+	name            string
+}{
+	{true, true, VersionTLS13, VersionTLS13, true, true, "tls13: DC server and client support"},
+	{true, false, VersionTLS13, VersionTLS13, true, false, "DC not server support"},
+	{false, true, VersionTLS13, VersionTLS13, true, true, "DC not client support"},
```
cjpatton: Here again: if one endpoint indicates support but the other doesn't, then DCs shouldn't be used in the connection.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -709,6 +709,34 @@ func certificateRequestInfo(certReq *certificateRequestMsgTLS13, vers uint16) *C
 	return cri
 }
 
+// getClientDelegatedCredential will return a Delegated Credential pair (a
+// Delegated Credential and its private key) for the given CertificateRequestInfo,
+// defaulting to the first element of cert.DelegatedCredentialPair.
+// The returned Delegated Credential could be invalid for usage in the handshake.
```
cjpatton, line 715:
```suggestion
// The returned Delegated Credential could be invalid for usage in the handshake.
// Returns an error if and only if a delegated credential cannot be configured for the
// current connection.
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -445,6 +445,34 @@ func cloneHash(in hash.Hash, h crypto.Hash) hash.Hash {
 	return out
 }
 
+// getDelegatedCredential will return a Delegated Credential pair (a Delegated
+// Credential and its private key) for the given ClientHelloInfo, defaulting to
+// the first element of cert.DelegatedCredentialPair.
+// The returned Delegated Credential could be invalid for usage in the handshake.
```
cjpatton, line 451:
```suggestion
// The returned Delegated Credential could be invalid for usage in the handshake.
// Returns an error if and only if a delegated credential cannot be configured for the 
// current connection.
```

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -723,26 +751,17 @@ func (hs *clientHandshakeStateTLS13) sendClientCertificate() error {
...
-		}
-
-		if dc != nil && priv != nil {
-			cert.PrivateKey = priv
-			if dc.raw == nil {
-				dc.raw, err = dc.marshal()
-				if err != nil {
-					c.sendAlert(alertInternalError)
-					return err
+	var dcStruct *DelegatedCredentialPair
```
cjpatton: Suggest renaming `dcStruct` to `dcPair`, Here and elsewhere.

cjpatton (commented):
One final question from me: From the tests it appears as if a DC may still be used in the connection even if support for DCs was not indicated by the peer. Is this an artifact of the test code, or is this what's actually happening?

src/crypto/tls/delegated_credentials_test.go
claucece: This is expected behaviour. 

This is how it works for server auth:

1. Client sends support for DC in its client hello
2. Server receives the client hello, sees that client supports DC.
3. Server checks in storage that it has DCs
4. Servers sends a DC

There is no need for the server to advertise support of DCs. That flag is only set when you are advertising the other party that you will like to use a DC (and, therefore, please send me one).

src/crypto/tls/delegated_credentials_test.go
claucece: This is expected behaviour. 

This is how it works for client auth:

1. Server sends support for DC in its certificate request
2. Client receives the certificate request, sees that server supports DC.
3. Client checks in storage that it has DCs
4. If DCs are found, Client sends a DC

src/crypto/tls/delegated_credentials_test.go
cjpatton: Resolution: the serverDCAuth flag has no impact for the server auth tests. We'll just remove it. (Similar change for the client uath tests.)

claucece:
> If I understand correctly, the point of this change is to provide the DC in a single callback via GetCertificate rather than two (GetCertificate followed by GetDelegatedCredential)?
> Could you include the motivation for the change in the commit message? Thanks, it looks good with minor comments!

Yes, the idea is to unify into single callback. I'll include in the commit message ;)

src/crypto/tls/delegated_credentials_test.go
claucece: Changed now ;)

src/crypto/tls/delegated_credentials_test.go
claucece: It is more for testing with each specific DC and algorithm (a DC with ed25519, a Dc with eddsa256, etc). I'll prefer if the tests were only using one DC (it simplifies all) but is was requested by reviewers to test with them all.

src/crypto/tls/handshake_client_tls13.go
claucece: Added.

src/crypto/tls/handshake_server_tls13.go
claucece: Added.

src/crypto/tls/handshake_client_tls13.go
claucece: Added.

cjpatton (approved):
LGTM

src/crypto/tls/common.go
```diff
@@ -1462,6 +1458,16 @@ func (c *Config) writeKeyLog(label string, clientRandom, secret []byte) error {
 // and is only for debugging, so a global mutex saves space.
 var writerMutex sync.Mutex
 
+// A DelegatedCredentialPair contains a Delegated Credential and its
+// associated private key.
+type DelegatedCredentialPair struct {
+	// DC is the delegated credential.
+	DC *DelegatedCredential
+	// PrivateKey is the private key corresponding to the public key of
+	// the DC. This must implement crypto.Signer.
```
armfazh:
```suggestion
	// PrivateKey is the private key used to derive the public key
	// contained in the DC. PrivateKey must implement crypto.Signer.
```

src/crypto/tls/common.go
```diff
@@ -1479,9 +1485,15 @@ type Certificate struct {
 	// SignedCertificateTimestamps contains an optional list of Signed
 	// Certificate Timestamps which will be served to clients that request it.
 	SignedCertificateTimestamps [][]byte
-	// DelegatedCredential is a serialized Delegated Credential, signed by
-	// the leaf certificate.
-	// If not supported, this will be nil.
+	// DelegatedCredentials are a list of Delegated Credentials with their
+	// corresponding private keys, signed by the leaf certificate.
+	// If not supported, this field will be nil.
```
armfazh:
```suggestion
	// If delegated credentials are not supported, this field is nil.
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -445,6 +445,36 @@ func cloneHash(in hash.Hash, h crypto.Hash) hash.Hash {
...
+// Returns an error if there are no delegated credentials or if the one found
+// cannot be used for the current connection.
+func getDelegatedCredential(clientHello *ClientHelloInfo, cert *Certificate) (*DelegatedCredentialPair, error) {
+	if len(cert.DelegatedCredentials) == 0 {
+		return nil, errors.New("No valid Delegated Credential found.")
```
armfazh:
```suggestion
		return nil, errors.New("No Delegated Credential found.")
```

armfazh: to output the same error as before

armfazh (approved):
few nits

src/crypto/tls/common.go
claucece: Changed.

src/crypto/tls/common.go
claucece: Changed.

src/crypto/tls/handshake_server_tls13.go
claucece: Changed.

src/crypto/tls/delegated_credentials_test.go
Lekensteyn: So based on a private chat with Sofía, my understanding is that it deliberately includes more than one to test this code path:
https://github.com/cloudflare/go/blob/1777b4c8ee76d5bc3e230d21fe3539d9cb565e33/src/crypto/tls/handshake_server_tls13.go#L464-L472

Lekensteyn (approved):
With an updated commit message with the motivation, it looks good to me.

claucece:
Commit message has been changed. Merging now.

PR #70, opened 2021-03-24
crypto/tls: NewDelegatedCredential returns a delegatedCredentialsPair.

armfazh:
Uses the `delegatedCredentialsPair` type for returning a DC.

### DIFF
--- a/src/crypto/tls/delegated_credentials_test.go
+++ b/src/crypto/tls/delegated_credentials_test.go
@@ -237,17 +237,17 @@ func initDCTest() {
 	}
 
 	for i := 0; i < len(dcTestDCSignatureScheme); i++ {
-		dc, priv, err := NewDelegatedCredential(dcCertP256, dcTestDCSignatureScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, false)
+		dc, err := NewDelegatedCredential(dcCertP256, dcTestDCSignatureScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, false)
 		if err != nil {
 			panic(err)
 		}
-		serverDC = append(serverDC, DelegatedCredentialPair{dc, priv})
+		serverDC = append(serverDC, *dc)
 
-		dc, priv, err = NewDelegatedCredential(dcCertP256, dcTestDCSignatureScheme[i], dcNow.Sub(dcCertP256.Leaf.NotBefore)+dcMaxTTL, true)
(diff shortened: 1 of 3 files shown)


claucece:
I thought of this a lot, and the reason why I didn't want to do it is because a storage mechanism might want to store the private key separated from the delegated credential: you will have to fetch the private key from the 'secure' storage and associated to the DC to configure TLS.

This thinking follows the same pattern of the Certificate: when a certificate is created, the private key is storage separately from it. So, I'll rather not have this change.

It is easy to create this structure outside of the Golang library itself, and then pass it to the configuration. This follows the same mechanism that when you assign a Certificate, you pass it separately the private key.

armfazh:
it makes sense, so the caller do not to mix private and public in the same struct.

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -293,10 +293,10 @@ func TestDelegateCredentialsValidate(t *testing.T) {
 	cert := dcTestCerts["dcP384"]
 	validTime := dcNow.Sub(cert.Leaf.NotBefore) + dcMaxTTL
 
-	delegatedCred, _, err := NewDelegatedCredential(cert, ECDSAWithP384AndSHA384, validTime, false)
+	delegatedCred, err := NewDelegatedCredential(cert, ECDSAWithP384AndSHA384, validTime, false)
 	if err != nil {
 		t.Fatal(err)
-	} else if delegatedCred == nil {
+	} else if delegatedCred.DC == nil {
```
thibmeu, line 299: nits: `else` is not needed here.
`Fatal` includes a `FailNow`. See [godoc](https://golang.org/pkg/testing/#B.Fatal)

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -307,68 +307,68 @@ func TestDelegateCredentialsValidate(t *testing.T) {
...
 
 	// Test DC with another certificate
 	cert = dcTestCerts["dcP521"]
 	validTime = dcNow.Sub(cert.Leaf.NotBefore) + dcMaxTTL
-	delegatedCred, _, err = NewDelegatedCredential(cert, ECDSAWithP384AndSHA384, validTime, false)
+	delegatedCred, err = NewDelegatedCredential(cert, ECDSAWithP384AndSHA384, validTime, false)
 	if err != nil {
 		t.Fatal(err)
-	} else if delegatedCred == nil {
+	} else if delegatedCred.DC == nil {
```
thibmeu, line 366: same here, `else` not needed

Issue #66, opened 2021-03-15
Update KEMTLS code


claucece:
Given that dc are now of the `cf` branch, KEMTLS can be implemented there as well. The KEMTLS functionality is implemented in the `cf-pq-kemtls` branch but needs to be updated to the latest DC code. 

The task consists of:

- [x] Updating KEMTLS code to the Go 1.16
- [x] Updating KEMTLS code to the latest DC code
- [x] Creating a parameter that allows the usage of KEMTLS and acts as a barrier so when it is set to false it cannot be used (for security purposes of TLS-1000).

claucece:
cc./ @armfazh @thomwiggers @Lekensteyn

Issue #68, opened 2021-03-15
Add hybrid KEMTLS


claucece:
We need to add a version of a hybrid KEMTLS. This needs to:

- [ ] Define an specification for how Hybrid KEMTLS works
- [ ] Add the appropriate changes to the KEMTLS code.

Depends on #66

claucece:
cc./ @armfazh @bwesterb @thomwiggers

thomwiggers:
"hybrid KEMTLS" is just KEMTLS with hybrid KEMs for the ephemeral and static (leaf cert) KEM, and and hybrid signature schemes in the signature chain.

The security proof does not rely on fancy properties of the KEMs or signature schemes, so we don't need to make any protocol changes in that sense. We might need to pick some identifiers but that's not any different from what needs to be done for PQ-only KEMs.

This is probably most easily implemented by treating the hybrid KEM as a completely independent algorithm from its constituent components. Treat "KyberX25519" is a "proper" algorithm and do not give it any special treatment on the wire compared to "X25519" or "Kyber". If that means duplicate X25519 public keys: so be it.

So that means following the same strategy as is proposed for [draft-ietf-tls-hybrid-design](https://datatracker.ietf.org/doc/html/draft-ietf-tls-hybrid-design). This draft only covers the (ephemeral) key exchange, of course, so we need to adjust also for the static KEM key exchange: just slap two keys together in the certificate as well and do the same for the ciphertext sent back to the server/client.

For certificates we can probably also do something similar: just slap two signatures together and verify both.

thomwiggers:
So my previous post covers the specification more-or-less 😉 

With regards to implementation changes that'd then be:

* [ ] Add hybrid algorithms
* [ ] Add algorithms to negotiation (i.e. supported algorithm lists)

bwesterb:
I agree with Thom.  Circl already supports some hybrid KEMs (eg. `Kyber512-X25519`) and Signatures (eg. `Ed25519-Dilithium3`.). For the latter we have also have OIDs (see `Old()` in Circl.)  For the former not yet — Cloudflare can just register them though — it's on the internal wiki how to.

We also need internal TLS identifiers, but we can just pick some in the reserved space.  See `TLSIdentifier` in Circl.

claucece:
yes! thanks you @thomwiggers @bwesterb 

> For the former not yet — Cloudflare can just register them though — it's on the internal wiki how to.

I'll check about this. 

[...]

PR #71, opened 2021-03-30
cryto/tls: Implement kemtls with mutual auth #66

claucece:
Ready for review.


src/crypto/kem/kem.go
```diff
@@ -0,0 +1,212 @@
...
+	"golang.org/x/crypto/cryptobyte"
+	"golang.org/x/crypto/curve25519"
+)
+
+// ID identifies each type of KEM.
+type ID uint16
+
+const (
+	// KEM25519 is X25519 as a KEM. Not quantum-safe.
+	KEM25519 ID = 0x01fb
```
wbl, line 19: Look at how this library implements hash functions to have a properly object oriented and extensible interface.

src/crypto/kem/kem.go
```diff
@@ -0,0 +1,212 @@
...
+func (pubKey *PublicKey) UnmarshalBinary(raw []byte) error {
+	s := cryptobyte.String(raw)
+
+	var id uint16
+	if !s.ReadUint16(&id) {
+		return errors.New("crypto/kem: invalid algorithm")
+	}
+
+	kemID := ID(id)
+	if kemID < minKEM || kemID > maxKEM {
```
wbl: This seems to imply we'll always have a single range of KEM support.

src/crypto/kem/kem.go
```diff
@@ -0,0 +1,212 @@
...
+		publicKey.Export(pubBytes)
+		privateKey.Export(privBytes)
+		return &PublicKey{KEMId: kemID, PublicKey: pubBytes}, &PrivateKey{KEMId: kemID, PrivateKey: privBytes}, nil
+	default:
+		return nil, nil, fmt.Errorf("crypto/kem: internal error: unsupported KEM %d", kemID)
+	}
+}
+
+// Encapsulate returns a shared secret and a ciphertext.
+func Encapsulate(rand io.Reader, pk *PublicKey) ([]byte, []byte, error) {
```
wbl: These return values might want names

src/crypto/tls/auth.go
```diff
@@ -267,39 +272,50 @@ func signatureSchemesForCertificate(version uint16, cert *Certificate) []Signatu
...
 func signatureSchemeForDelegatedCredential(version uint16, dc *DelegatedCredential) []SignatureScheme {
 	pub := dc.cred.publicKey
-
 	var sigAlgs []SignatureScheme
-	switch pub.(type) {
-	case *ecdsa.PublicKey:
-		pk, ok := pub.(*ecdsa.PublicKey)
-		if !ok {
+
+	kemPub, ok := pub.(*kem.PublicKey)
```
wbl: Why can't this be part of the switch statement?

src/crypto/tls/common.go
```diff
@@ -1135,9 +1178,14 @@ func supportedVersionsFromMax(maxVersion uint16) []uint16 {
 }
 
 var defaultCurvePreferences = []CurveID{X25519, CurveP256, CurveP384, CurveP521}
+var defaultKEMPreferences = []CurveID{SIKEp434, Kyber512}
 
 func (c *Config) curvePreferences() []CurveID {
 	if c == nil || len(c.CurvePreferences) == 0 {
+		if c.AllowKEMTLS {
```
wbl: What happens when a client has AllowKEMTLS on, but the server doesn't speak it?

src/crypto/tls/delegated_credentials.go
```diff
@@ -364,6 +366,16 @@ func NewDelegatedCredential(cert *Certificate, pubAlgo SignatureScheme, validTim
 		if err != nil {
 			return nil, nil, err
 		}
+	case KEMTLSWithSIKEp434:
+		pubK, privK, err = kem.GenerateKey(rand.Reader, kem.SIKEp434)
```
wbl, line 379: These cases could be combined via the TLS to KEM number mapping

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -486,7 +503,7 @@ func testClientGetCertificate(cr *CertificateRequestInfo) (*Certificate, error)
 
 // Tests the handshake and one round of application data. Returns true if the
 // connection correctly used a Delegated Credential.
-func testConnWithDC(t *testing.T, clientMsg, serverMsg string, clientConfig, serverConfig *Config, peer string) (bool, error) {
+func testConnWithDC(t *testing.T, clientMsg, serverMsg string, clientConfig, serverConfig *Config, peer string, kemtls bool) (bool, bool, error) {
```
wbl: What is the second boolean return argument?

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -644,3 +671,194 @@ func TestDCHandshakeClientAndServerAuth(t *testing.T) {
...
+	clientConfig := dcTestConfig.Clone()
+	clientConfig.GetClientCertificate = testClientGetKEMCertificate
+	clientConfig.SupportDelegatedCredential = true // to force the sending of a KEM DC
+	clientConfig.AllowKEMTLS = true
+	clientConfig.CurvePreferences = []CurveID{SIKEp434, Kyber512}
+
+	for i, test := range dcKEMClientTests {
+		serverConfig.SupportDelegatedCredential = test.serverDCSupport
+
+		for dcCount = 0; dcCount < len(dcTestDCKEMScheme); dcCount++ {
```
wbl, line 834: I realize dcount escaped the first time, but  I don't like the use of a global in tests, especially when we can use the configuration options instead.

src/crypto/tls/delegated_credentials_test.go
```diff
@@ -644,3 +671,194 @@ func TestDCHandshakeClientAndServerAuth(t *testing.T) {
...
+	serverMsg := "hello, client"
+
+	serverConfig := dcTestConfig.Clone()
+	serverConfig.ClientAuth = RequestClientKEMCert
+	serverConfig.GetCertificate = testServerGetKEMCertificate
+	clientConfig := dcTestConfig.Clone()
+	clientConfig.GetClientCertificate = testClientGetKEMCertificate
+	clientConfig.AllowKEMTLS = true
+	serverConfig.AllowKEMTLS = true
+	clientConfig.CurvePreferences = []CurveID{Kyber512}
```
wbl, line 907: Doesn't this have to change with the kems?

src/crypto/tls/handshake_client.go
```diff
@@ -119,23 +123,65 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
...
-		curveID := config.curvePreferences()[0]
-		if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
-			return nil, nil, errors.New("tls: CurvePreferences includes unsupported curve")
-		}
-		params, err = generateECDHEParameters(config.rand(), curveID)
-		if err != nil {
-			return nil, nil, err
+		if !config.AllowKEMTLS {
+			curveID := config.curvePreferences()[0]
+			if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
```
wbl, line 231: Why is this curve special?

src/crypto/tls/handshake_client.go
```diff
@@ -119,23 +123,65 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
...
+			params, err = generateECDHEParameters(config.rand(), curveID)
+			if err != nil {
+				return nil, nil, err
+			}
+			keyShares = append(keyShares, keyShare{group: curveID, data: params.PublicKey()})
+			keySharePrivates = append(keySharePrivates, params)
+		} else {
+			var haveECDHE, haveKEM bool
+
+			// loop over supported curves and kems until there is a KEM and an ECDHE curve
```
wbl, line 244: Why not do the hybrid in the kems?

src/crypto/tls/handshake_client_kemtls.go
```diff
@@ -0,0 +1,289 @@
...
+		return err
+	}
+
+	if hs.certReq.supportDelegatedCredential && len(hs.certReq.supportedSignatureAlgorithmsDC) > 0 {
+		var dcPair *DelegatedCredentialPair
+		if delegatedCredentialPair, err := getClientDelegatedCredential(cri, cert); err == nil {
+			if delegatedCredentialPair.DC != nil && delegatedCredentialPair.PrivateKey != nil {
+				var err error
+				// Even if the Delegated Credential has already been marshalled, be sure it is the correct one.
+				if delegatedCredentialPair.DC.raw, err = delegatedCredentialPair.DC.marshal(); err == nil {
```
wbl, line 167: Shouldn't this be handled inside of marshal?

src/crypto/tls/handshake_client_kemtls.go
```diff
@@ -0,0 +1,289 @@
...
+				}
+			}
+		}
+	}
+
+	_, ok := cert.PrivateKey.(*kem.PrivateKey)
+	if !ok {
+		// it has to be a KEM key
+		c.sendAlert(alertInternalError)
+		return nil
```
wbl: Why not returning an error here?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -280,21 +292,48 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
...
+			kemID := kem.ID(curveID)
+			pk, sk, err := kem.GenerateKey(c.config.rand(), kemID)
+			if err != nil {
+				c.sendAlert(alertInternalError)
+				return errors.New("tls: CurvePreferences includes unsupported kem")
+			}
+			hs.keyShare = []clientKeySharePrivate{sk}
+			hs.helloBase.keyShares = []keyShare{{group: CurveID(pk.KEMId), data: pk.PublicKey}}
+		} else {
+			if _, ok := curveForCurveID(curveID); (curveID != X25519 && !curveID.isKEM()) && !ok {
```
wbl, line 329: why is X25519 special?

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -297,17 +311,29 @@ GroupSelection:
...
-		c.sendAlert(alertInternalError)
-		return err
+	if selectedGroup.isKEM() && c.config.AllowKEMTLS {
+		sharedKey, ciphertext, err := kem.Encapsulate(c.config.rand(), &kem.PublicKey{KEMId: kem.ID(selectedGroup), PublicKey: clientKeyShare.data})
+		if err != nil {
+			c.sendAlert(alertInternalError)
+			return errors.New("tls: CurvePreferences includes unsupported curve")
+		}
+		hs.hello.serverShare = keyShare{group: selectedGroup, data: ciphertext}
+		hs.sharedKey = sharedKey
```
wbl, line 431: So this confused me a bit. We've changed the meaning of these messages so that this is just a key the server has picked, and it's not until the client gets backed in the finished that we are done.

wbl (changes requested):
I think this is on the right track, but more needs to be done. it would be nice there was something describing what changes and why. Ideally the handling of ECDH and KEMS would be as unified as possible, but there are limits.

src/crypto/tls/handshake_client.go
claucece: This is how Golang is doing it prior to adding KEMTLS, so this is not a change introduced on this PR.

The reason why they do it is: `curveForCurveID` returns the elliptic.Curve given an eddsa id. 
`x25519` is not part of that interface, and that is why later it is calculated as:

```
func generateECDHEParameters(rand io.Reader, curveID CurveID) (ecdheParameters, error) {
	if curveID == X25519 {
		privateKey := make([]byte, curve25519.ScalarSize)
		if _, err := io.ReadFull(rand, privateKey); err != nil {
			return nil, err
		}
		publicKey, err := curve25519.X25519(privateKey, curve25519.Basepoint)
		if err != nil {
			return nil, err
		}
		return &x25519Parameters{privateKey: privateKey, publicKey: publicKey}, nil
	}
```

src/crypto/tls/handshake_client.go
claucece: There is still not a consensus on how we are going to do hybrid kem tls. That will come on a PR that solves this: #68

src/crypto/tls/handshake_client_kemtls.go
claucece: This is how Golang is doing it prior to adding KEMTLS, so this is not a change introduced on this PR.

The reason why they do it is: `curveForCurveID` returns the elliptic.Curve given an eddsa id. 
`x25519` is not part of that interface, and that is why later it is calculated as:

```
func generateECDHEParameters(rand io.Reader, curveID CurveID) (ecdheParameters, error) {
	if curveID == X25519 {
		privateKey := make([]byte, curve25519.ScalarSize)
		if _, err := io.ReadFull(rand, privateKey); err != nil {
			return nil, err
		}
		publicKey, err := curve25519.X25519(privateKey, curve25519.Basepoint)
		if err != nil {
			return nil, err
		}
		return &x25519Parameters{privateKey: privateKey, publicKey: publicKey}, nil
	}
```

src/crypto/tls/handshake_client_tls13.go
claucece: This is how Golang is doing it prior to adding KEMTLS, so this is not a change introduced on this PR.

The reason why they do it is: `curveForCurveID` returns the elliptic.Curve given an eddsa id. 
`x25519` is not part of that interface, and that is why later it is calculated as:

```
func generateECDHEParameters(rand io.Reader, curveID CurveID) (ecdheParameters, error) {
	if curveID == X25519 {
		privateKey := make([]byte, curve25519.ScalarSize)
		if _, err := io.ReadFull(rand, privateKey); err != nil {
			return nil, err
		}
		publicKey, err := curve25519.X25519(privateKey, curve25519.Basepoint)
		if err != nil {
			return nil, err
		}
		return &x25519Parameters{privateKey: privateKey, publicKey: publicKey}, nil
	}
```

src/crypto/kem/kem.go
claucece: Yeah.. I'll remove this ;)

src/crypto/kem/kem.go
claucece: Added ;)

src/crypto/tls/auth.go
claucece: Oh, I overlooked this. Thank you for pointing it out ;)

src/crypto/tls/common.go
claucece: This is a really good question. I added the `defaultCurvePreferences` to the `defaultKEMPreferences` as well, just to add fallback algorithms.

src/crypto/tls/delegated_credentials.go
claucece: I agree. I'll leave it as it is, as this seems to be a consideration for the draft itself. I'll add an issue on the draft repo.

src/crypto/tls/delegated_credentials_test.go
claucece: It states if the connection use kemtls. I'll add names to the return arguments to make it clearer.

src/crypto/tls/delegated_credentials_test.go
claucece: Yeah, it is using `Kyber512`.

src/crypto/tls/handshake_client_kemtls.go
claucece: It is sort of expected that the raw is already marshaled. Same happens with the cert.

claucece: Good point! Adding it ;)

src/crypto/tls/handshake_client_kemtls.go
claucece: Good point! Adding it ;)

src/crypto/tls/handshake_server_tls13.go
claucece: Yes!

claucece:
> I think this is on the right track, but more needs to be done. it would be nice there was something describing what changes and why. Ideally the handling of ECDH and KEMS would be as unified as possible, but there are limits.

I agree. Let's look at this as the first implementation (poc) of draft 00 or pre-draft. Things might get clearer once a first draft is around.

src/crypto/kem/kem.go
claucece: I agree. I'll add it later today ;)

src/crypto/kem/kem.go
```diff
@@ -0,0 +1,203 @@
+package kem
```
cjpatton, line 1: AFAICT this package, `crypto/kem`, is simply a wrapper around `circl/kem/`. Why not call `circl/kem` from `crypto/tls` directly? It would be nice to not have to maintain a new package.

src/crypto/tls/testdata/Client-TLSv10-ClientCert-ECDSA-ECDSA
```diff
@@ -17,11 +17,11 @@
...
-00000000  16 03 01 00 59 02 00 00  55 03 01 09 68 1e 95 bd  |....Y...U...h...|
-00000010  f3 a3 32 97 d0 26 e7 c7  a5 5c b2 31 6c e2 07 41  |..2..&...\.1l..A|
-00000020  95 59 75 18 52 a8 7d 77  5a 1e be 20 35 f2 d1 f7  |.Yu.R.}wZ.. 5...|
-00000030  cf 70 2a 47 5a 5c bb 36  51 4c 5e f4 82 f4 db 12  |.p*GZ\.6QL^.....|
-00000040  85 7f d6 14 64 f2 e0 68  cd d0 60 1a c0 09 00 00  |....d..h..`.....|
+00000000  16 03 01 00 59 02 00 00  55 03 01 27 14 34 e9 ef  |....Y...U..'.4..|
+00000010  89 88 53 1c 48 31 03 98  8c 7c 56 63 bc 6b 2c 86  |..S.H1...|Vc.k,.|
+00000020  63 39 d2 0f 1d a0 b8 6b  c5 11 84 20 43 83 47 5b  |c9.....k... C.G[|
+00000030  ab bb 57 2a 30 7d 37 3f  a1 44 56 87 fb 14 33 8d  |..W*0}7?.DV...3.|
+00000040  72 cb 79 61 ee 4b fd 83  66 e8 e7 a2 c0 09 00 00  |r.ya.K..f.......|
```
cjpatton: Is regenerating the test vectors strictly needed for this PR? If not, would you mind reverting these changes and submitting them in a different PR?

src/crypto/tls/common.go
```diff
@@ -125,8 +126,18 @@ const (
 	CurveP384 CurveID = 24
 	CurveP521 CurveID = 25
 	X25519    CurveID = 29
+	SIKEp434  CurveID = CurveID(kem.SIKEp434)
+	Kyber512  CurveID = CurveID(kem.Kyber512)
```
cjpatton, line 132: Since these code points are used only in KEMTLS, I would prefix the variables with `KEMTLS`. (I.e., `KEMTLSSKEp434` and `KEMTLSKyber512`.)

src/crypto/tls/common.go
```diff
@@ -375,6 +395,10 @@ const (
 	// during the handshake, but does not require that the client send any
 	// certificates.
 	RequestClientCert
+	// RequestClientKEMCert indicates that a client certificate with a KEM
+	// public key should be requested during the handshake, but does not
+	// require that the client send any certificates.
+	RequestClientKEMCert
```
cjpatton: Is adding a new `ClientAuthType` variant necessary? Here's where this variant is used in `handshake_server_kemtls.go`:
```go
   func (hs *serverHandshakeStateTLS13) requestClientKEMCert() bool {
      return hs.c.config.ClientAuth >= RequestClientKEMCert && !hs.usingPSK                                                                                                                                
   }     
```
It seems like you could use `RequestClientCert` here instead. The only other place this is used is in `common_string.go`, but there it seems like you could do something else.

The reason I bring it up is that this makes the API confusing, since "KEMCert" only has meaning in for KEMTLS.

src/crypto/tls/common.go
```diff
@@ -466,8 +490,22 @@ const (
 	// Legacy signature and hash algorithms for TLS 1.2.
 	PKCS1WithSHA1 SignatureScheme = 0x0201
 	ECDSAWithSHA1 SignatureScheme = 0x0203
+
+	// KEMTLS algorithms for the Post-Quantum Cryptography experiment.
+	// NOTE: Do not use outside of the experiment.
+	KEMTLSWithSIKEp434 SignatureScheme = 0xfe00
+	KEMTLSWithKyber512 SignatureScheme = 0xfe01
```
cjpatton, line 516:
```suggestion
	KEMTLSSIKEp434 SignatureScheme = 0xfe00
	KEMTLSKyber512 SignatureScheme = 0xfe01
```
The "With" doesn't seem necessary.

src/crypto/tls/common.go
```diff
@@ -824,6 +862,10 @@ type Config struct {
 	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
 	SupportDelegatedCredential bool
 
+	// AllowKEMTLS is true if the client or server is willing
+	// to start a KEMTLS handshake based on TLS 1.3.
+	AllowKEMTLS bool
```
cjpatton:
```suggestion
	KEMTLSEnabled bool
```
This aligns better with other flags in `Config` (e.g., `ECHEnabled`).

src/crypto/tls/common.go
```diff
@@ -1135,9 +1178,14 @@ func supportedVersionsFromMax(maxVersion uint16) []uint16 {
 }
 
 var defaultCurvePreferences = []CurveID{X25519, CurveP256, CurveP384, CurveP521}
+var defaultKEMPreferences = []CurveID{SIKEp434, Kyber512, X25519, CurveP256, CurveP384, CurveP521}
```
cjpatton:
```suggestion
var defaultKEMTLSPreferences = []CurveID{SIKEp434, Kyber512, X25519, CurveP256, CurveP384, CurveP521}
```
This will make it easier to find this variable.

Also, I'd remove the NIST curves unless we plan to use them.

src/crypto/tls/handshake_client.go
```diff
@@ -119,23 +123,65 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
...
 	if hello.supportedVersions[0] == VersionTLS13 {
 		hello.cipherSuites = append(hello.cipherSuites, defaultCipherSuitesTLS13()...)
 
-		curveID := config.curvePreferences()[0]
-		if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
-			return nil, nil, errors.New("tls: CurvePreferences includes unsupported curve")
-		}
-		params, err = generateECDHEParameters(config.rand(), curveID)
-		if err != nil {
-			return nil, nil, err
```
cjpatton, line 129: Can the following code be reworked so that we don't have to write down this code twice? It would make it easier to read and maintain.
```go
              curveID := config.curvePreferences()[0]
              if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
                  return nil, nil, errors.New("tls: CurvePreferences includes unsupported curve")
              }        
              params, err = generateECDHEParameters(config.rand(), curveID)
              if err != nil {
                  return nil, nil, err
              }        
              keyShares = append(keyShares, keyShare{group: curveID, data: params.PublicKey()})
              keySharePrivates = append(keySharePrivates, params)
``

File: src/crypto/tls/handshake_client.go [hunk 24]
```diff
@@ -119,23 +123,65 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
...
+
+				if haveECDHE && haveKEM {
+					break
+				}
+			}
 		}
-		hello.keyShares = []keyShare{{group: curveID, data: params.PublicKey()}}
+
+		hello.keyShares = keyShares
 		hello.delegatedCredentialSupported = config.SupportDelegatedCredential
```
cjpatton, line 278: What happens if `AllowKEMTLS` is set but `SupportDelegatedCredential` is not?

claucece: There is not problem as a certificate with a kem is also allowed.

src/crypto/kem/kem.go
claucece: It is not a one to one wrapper. What this package does is that it exposes an interface that is used by tls based on the kem functions. I can move this code to its own file on the tls folder; but it seemed a little weird to have it there.

src/crypto/tls/common.go
claucece: The reason why it was added was to constrain usage. But sure, I can use the other one.

src/crypto/tls/common.go
claucece: Sure.

src/crypto/tls/common.go
claucece: The NIST curves are the fallback algorithms; so yes, they could be used if you support some kem but the peer did not. 

> unless we plan to use them

what does this mean?

src/crypto/tls/testdata/Client-TLSv10-ClientCert-ECDSA-ECDSA
claucece: Yes, it is needed. All the tests fails if they are not re-generated. I'm not sure why is it needed to review the test vectors instead of just ignoring these files.. so is there a reason why you are reviewing them?

src/crypto/tls/handshake_client.go
claucece: Sure. I'll try refactoring it.

src/crypto/tls/common.go
claucece: I disagree on this one. These are the kem algorithms that can used in combination, for example, with pq signatures (which will be another experiment we will run), in which kem are used for confidentiality. So they are not "kemtls" per se ;)

src/crypto/tls/common.go
claucece: Sure!

src/crypto/kem/kem.go
cjpatton: What about moving it `crypto/kem` to `crypto/tls/internal/kem`?

src/crypto/tls/testdata/Client-TLSv10-ClientCert-ECDSA-ECDSA
cjpatton: Not necessary, just trying to find ways to reduce the scope. Thanks for clarifying, you can disregard this comment.

src/crypto/tls/common.go
cjpatton: Hmm, in retrospect I think this comment didn't make much sense. What I meant was: "KEMTLS clients don't need to advertise support for these curves". Looking at the code closer, I realized that the may end up falling back to these curves if the server doesn't support KEMTLS.

Disregard the comment about the NIST curves. I still think we should change the name, however.

src/crypto/tls/common.go
cjpatton: True, but they're also not actually "curves". KEMTLS is already overloading the semantics of the `CurveID` by adding codepoints for KEMs. To make this fact clear, I would at least add the prefix "KEM" to the variables. I think "KEMTLS" is better because it makes it clear that these code points are used to negotiate KEMTLS and not TLS.

cjpatton: Another advantage is that it's easier to search through the code for variables related to KEMTLS.

src/crypto/tls/handshake_client.go
cjpatton: But KEMTLS couldn't be negotiated right? Since the client isn't advertising support for DCs?

BTW, I'm not saying anything needs to be changed here. I'm just trying to understand how KEMTLS interacts with other extensions.

src/crypto/tls/handshake_client_kemtls.go
cjpatton: Yeah, this is a thing about this particular package. It's a little weird, but @claucece is following the package's code style here.

src/crypto/kem/kem.go
cjpatton: As codeowner for `crypto/tls`, you would be codeowner of this internal package. Otherwise, we would need to manually assign you as codeowner of `crypto/kem`.

src/crypto/tls/handshake_client.go
claucece: If AllowKEMTLS is set that means two things:

* A KEM will be used for key exchange
* a KEM will be used as a signature: either by having it on a DC or in a leaf certificate

KEMTLS is not restricted to DCs, so if the SupportDelegatedCredential is not set to true, it is expected to fallback to a certificate that has a KEM. If a certificate with a KEM is not found or a DC with a KEM is not found, the handshake fails, even if a KEM was advertised and correctly negotiated for a key exchange.

src/crypto/kem/kem.go
claucece: > Otherwise, we would need to manually assign you as codeowner of crypto/kem.

And what is the problem with this?

claucece: Leaving this discussion for next next week.

src/crypto/tls/common.go
claucece: It is overloading for the moment as it is reusing an existing variable name of the draft. If later such a thing as `kemPreferences` exist is where the kems will exist. Overloading as, as it has been stated, this is a poc of concept of the kemtls idea and there is no draft still to be based upon.

These are post-quantum variables, not variables related to KEMTLS.

I'll leave this as a decision for the draft: if there will be such a thing as `kemPreferences`.

src/crypto/tls/handshake_client.go
claucece: Leaving this for next next week.

PR #73, opened 2021-03-31
crypto/tls: allow the usage of other keyUsage when checking for the d…

### LINKED ISSUE #72: crypto/tls: keyUsage tests exclusively for DigitalSignature flag
The `tls.Config.KeyUsage` field can contain several flags. 
Currently, for generating a delegated credential extension, it tests that the `KeyUsageDigitalSignature` flags is enabled exclusively. 

https://github.com/cloudflare/go/blob/5ef1b90573f8742b763b7e65a43ce7fa20e37bb4/src/crypto/x509/x509.go#L1789-L1795

Shouldn't this 
`(template.KeyUsage == KeyUsageDigitalSignature)` 
be replaced by 
`(template.KeyUsage &  KeyUsageDigitalSignature != 0)`

i.e., only testing the flag is present, without baning from setting other flags in the same field.

claucece:
Should fix #72

### DIFF
--- a/src/crypto/x509/x509.go
+++ b/src/crypto/x509/x509.go
@@ -1787,7 +1787,7 @@ func buildCertExtensions(template *Certificate, subjectIsEmpty bool, authorityKe
 	}
 
 	// This extension is not critical
-	if template.KeyUsage == KeyUsageDigitalSignature && template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) {
+	if template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) && (template.KeyUsage&KeyUsageDigitalSignature != 0) {
 		ret[n].Id = oidExtensionDelegatedCredential
 		ret[n].Value, err = asn1.Marshal(asn1.NullRawValue)
 		if err != nil {


src/crypto/x509/x509.go
```diff
@@ -1787,13 +1787,15 @@ func buildCertExtensions(template *Certificate, subjectIsEmpty bool, authorityKe
 	}
 
 	// This extension is not critical
-	if template.KeyUsage == KeyUsageDigitalSignature && template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) {
-		ret[n].Id = oidExtensionDelegatedCredential
-		ret[n].Value, err = asn1.Marshal(asn1.NullRawValue)
-		if err != nil {
-			return
+	if template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) {
+		if template.KeyUsage&KeyUsageDigitalSignature != 0 {
```
cjpatton: Why next this if statement? I think it's clearer if you move the condition into the if statement above.

cjpatton (changes requested):

src/crypto/x509/x509.go
claucece: Solved.

src/crypto/x509/x509.go
```diff
@@ -1787,7 +1787,7 @@ func buildCertExtensions(template *Certificate, subjectIsEmpty bool, authorityKe
 	}
 
 	// This extension is not critical
-	if template.KeyUsage == KeyUsageDigitalSignature && template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) {
+	if template.AllowDC && !template.IsCA && !oidInExtensions(oidExtensionDelegatedCredential, template.ExtraExtensions) && template.KeyUsage&KeyUsageDigitalSignature != 0 {
```
lukevalenta: Perhaps add some parentheses here to improve readability for the bit operation? `...&& (template.KeyUsage & KeyUsageDigitalSignature != 0) {`

lukevalenta (approved):
LGTM other than a comment to improve readability.

cjpatton (approved):

src/crypto/x509/x509.go
claucece: Added now.

PR #77, opened 2021-04-20
crypto/tls: Add cipher suite checks to ECH unit tests

### LINKED ISSUE #76: ECHConfig test cipher suites
It looks like the test code is exercising fewer HPKE cipher suites than it used to, and the change happened during the upgrade from draft-09 to draft-10. Perhaps the code in this repo should also assert on the length of the cipher suites list. I think it just loops.

Noticed this while working on Rustls:
https://github.com/grafica/rustls/blob/cde5c28faa03155350f51c49b8c6438391fc28bc/rustls/src/ech_key.rs#L189

cjpatton:
Closes #76.

Test that UnmarshalECHConfigs (resp. EXP_UnmarshalECHKeys) parses the
correct number of cipher suites.

### DIFF
--- a/src/crypto/tls/ech_test.go
+++ b/src/crypto/tls/ech_test.go
@@ -846,6 +846,16 @@ func TestUnmarshalConfigs(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	if len(configs) != 2 {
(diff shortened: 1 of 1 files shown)


claucece (approved):

claucece:
Thank you this!

PR #78, opened 2021-04-27
crypto/x509/pkix: Handle emailAddress OID

cjpatton:
Adds support for the emailAddress OID.

### DIFF
--- a/src/crypto/x509/pkix/pkix.go
+++ b/src/crypto/x509/pkix/pkix.go
@@ -24,6 +24,8 @@ type AlgorithmIdentifier struct {
 type RDNSequence []RelativeDistinguishedNameSET
 
 var attributeTypeNames = map[string]string{
+	"1.2.840.113549.1.9.1": "emailAddress",
+
 	"2.5.4.6":  "C",
 	"2.5.4.10": "O",
 	"2.5.4.11": "OU",
@@ -124,7 +126,7 @@ type Name struct {
 	Country, Organization, OrganizationalUnit []string
 	Locality, Province                        []string
 	StreetAddress, PostalCode                 []string
-	SerialNumber, CommonName                  string
+	SerialNumber, CommonName, EmailAddress    string
 
 	// Names contains all parsed attributes. When parsing distinguished names,
 	// this can be used to extract non-standard attributes that are not parsed
@@ -177,6 +179,10 @@ func (n *Name) FillFromRDNSequence(rdns *RDNSequence) {
 					n.PostalCode = append(n.PostalCode, value)
 				}
 			}
+
+			if t.Equal(asn1.ObjectIdentifier(oidEmailAddress)) {
+				n.EmailAddress = value
+			}
 		}
 	}
 }
@@ -191,6 +197,7 @@ var (
 	oidProvince           = []int{2, 5, 4, 8}
 	oidStreetAddress      = []int{2, 5, 4, 9}
 	oidPostalCode         = []int{2, 5, 4, 17}
+	oidEmailAddress       = []int{1, 2, 840, 113549, 1, 9, 1}
 )
 
 // appendRDNs appends a relativeDistinguishedNameSET to the given RDNSequence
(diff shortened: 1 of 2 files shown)


src/crypto/x509/pkix/pkix.go
```diff
@@ -177,6 +179,10 @@ func (n *Name) FillFromRDNSequence(rdns *RDNSequence) {
 					n.PostalCode = append(n.PostalCode, value)
 				}
 			}
+
+			if len(t) == 7 && t[0] == 1 && t[1] == 2 && t[2] == 840 && t[3] == 113549 && t[4] == 1 && t[5] == 9 && t[6] == 1 {
```
wbl: Following suite is fine, but why not a compareOID function?

wbl (approved):

src/crypto/x509/pkix/pkix.go
cjpatton: Good call, done.

Lekensteyn:
This limitation exists in upstream Go too. Earlier reports:

* https://github.com/golang/go/issues/15256 wanted to extract the `emailAddress` name. Right now they have to parse `certificate.Subject.Names` and search for the emailAddress OID.
* https://github.com/golang/go/issues/39924 also shows a case where they expected `emailAddress` to be human-readable. Closed since unknown attributes are now appended to the end.

Could you come up with any tests for this if it is important? If possible, you could also submit a proposal upstream.

Edit: RFC 5280 (https://tools.ietf.org/html/rfc5280#page-116) defines:
```
-- Legacy attributes
pkcs-9 OBJECT IDENTIFIER ::=
       { iso(1) member-body(2) us(840) rsadsi(113549) pkcs(1) 9 }
id-emailAddress      AttributeType ::= { pkcs-9 1 }
EmailAddress ::=     IA5String (SIZE (1..ub-emailaddress-length))
```

and also in https://tools.ietf.org/html/rfc5280#section-4.1.2.6

```
   Conforming implementations generating new certificates with
   electronic mail addresses MUST use the rfc822Name in the subject
   alternative name extension (Section 4.2.1.6) to describe such
   identities.  Simultaneous inclusion of the emailAddress attribute in
   the subject distinguished name to support legacy implementations is
   deprecated but permitted.
```

I think more work is needed if this field is intended to be fully supported for validation purposes. Is it possible to modify the user instead of stdlib?

cjpatton:
Thanks, @Lekensteyn, for providing some context for this change.

> Could you come up with any tests for this if it is important? If possible, you could also submit a proposal upstream.

Given (1) this feature is deprecated by RFC5280 and (2) it has been proposed and rejected before, I don't think we should try to upstream this change.

> I think more work is needed if this field is intended to be fully supported for validation purposes.

Can you be more specific about what changes are needed? Is it just that you want to see some tests, or are you concerned about the side-effects for users of crypto/x509/pkix?

> Is it possible to modify the user instead of stdlib?

This is certainly possible, but it would take some doing. To be clear, the goal is to replicate a feature of boringSSL that we need for our server: https://boringssl.googlesource.com/boringssl/+/refs/heads/master/include/openssl/nid.h#303

src/crypto/x509/pkix/pkix.go
```diff
@@ -260,6 +270,10 @@ func (n Name) String() string {
 					continue
 				}
 			}
+			if oidEqual(t, oidEmailAddress) {
```
Lekensteyn: There is already an implementation for this:
```suggestion
			if t.Equal(asn1.ObjectIdentifier(oidEmailAddress)) {
```

Lekensteyn:
> Can you be more specific about what changes are needed? Is it just that you want to see some tests, or are you concerned about the side-effects for users of crypto/x509/pkix?

Both, I was originally concerned about side-effects, but on a quick look there shouldn't be any except those that are strict about the `String()` output. Some tests would be nice to demonstrate the exact issue that you are trying to solve with this patch.

src/crypto/x509/pkix/pkix.go
cjpatton: Thanks!

src/crypto/x509/pkix/pkix_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestEmailAddress(t *testing.T) {
+	name := &Name{
+		Country:            []string{"US"},
+		Organization:       []string{"Hot Wheels LLC"},
+		OrganizationalUnit: []string{"Sweet Rides"},
+		Locality:           []string{"San Francisco"},
+		Province:           []string{"California"},
+		CommonName:         "test.com",
+		EmailAddress:       "hot@wheels.com",
```
Lekensteyn: In the rare case that the owner of this domain complains, can we use something such as `example.com` and `user@example.net`?

Lekensteyn (approved):
One comment about the chosen value in the tests, LGTM otherwise.

src/crypto/x509/pkix/pkix_test.go
cjpatton: Done.

PR #81, opened 2021-05-10
Generates the experiment name and stores it in the trace.

armfazh:


claucece:
I love this idea! Looking good for me!

claucece (approved):

Lekensteyn (approved):
Couldn't this information be obtained outside the stdlib as well without requiring changes to stdlib? LGTM otherwise

cjpatton:
Please change the commit message so that it has the right structure. Instead of
```
Includes the experiment name based on connection.
```

you want something like
```
crypto/tls: Include experiment name based on connection
```

Also, please add a description of what the change is and why it's needed to the body of the commit message. 

See https://github.com/cloudflare/go/wiki/Contributing

claucece:
@cjpatton while I think that might be necessary for our `cf` branch, this is wanting to be merged to `cf-pq-kemtls`, as seen in the request, which will be squashed and have different commits. Thefore, I don't think this needs to follow the commit convention.

cjpatton:
Ah yes, I didn't look at the base! Right you are, disregard.

PR #82, opened 2021-05-13
[cf-pq-kemtls] crypto/tls: reset timings

armfazh:
@claucece pls verify that the resets were inserted in the right position.
The idea is to have a reset inmediately after a peer has received a message from the other peer.

*Please review the order of the timings, mainy in the KEMTLS case.

See https://github.com/claucece/KEMTLS-local-measurements/pull/1

```sh
Client
|--> Write Client Hello       101.392µs 
   Server
   | Receive Client Hello     98.792µs 
   | Write Server Hello       131.945µs 
   | Write Server Enc Exts    153.087µs 
   | Write Server Certificate 165.049µs 
<--| Write Server Cert Verify 301.993µs 
Client
|    Process Server Hello       2.429µs 
|    Receive Server Enc Exts    77.393µs 
|    Receive Server Certificate 85.568µs 
|    Receive Server Cert Verify 477.781µs 
|    Receive Server Finished    479.148µs 
|    Write Client Certificate   0s 
|    Write Client Cert Verify   0s 
|--> Write Client Finished      515.741µs 
   Server
   | Receive Client Certificate 0s 
   | Receive Client Cert Verify 0s 
   | Receive Client Finished    549.793µs 
<--| Write Server Finished    309.795µs 
Client Total time: 1.044744ms 
Server Total time: 911.365µs
```

### DIFF
--- a/src/crypto/tls/handshake_client.go
+++ b/src/crypto/tls/handshake_client.go
@@ -352,6 +352,9 @@ func (c *Conn) clientHandshake() (err error) {
 		return err
 	}
 
+	// Client already received the serverHello
+	handshakeTimings.reset()
+
 	// If we are negotiating a protocol version that's lower than what we
 	// support, check for the server downgrade canaries.
 	// See RFC 8446, Section 4.1.3.
--- a/src/crypto/tls/handshake_client_kemtls.go
+++ b/src/crypto/tls/handshake_client_kemtls.go
@@ -55,7 +55,8 @@ func (hs *clientHandshakeStateTLS13) handshakeKEMTLS() error {
 		}
 	}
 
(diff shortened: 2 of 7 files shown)


claucece:
While I like this idea, it will be nice if we have also the possibility of checking the overall time of the handshake (without the resets). Can it be added as an option?

armfazh:
> While I like this idea, it will be nice if we have also the possibility of checking the overall time of the handshake (without the resets). Can it be added as an option?

Yes, this is already done in the `FullProtocol` metric, See timings at the end:
```
Client Total time: 1.044744ms 
Server Total time: 911.365µs
```

claucece:
Tested locally. Will send small fixes after merge.

PR #83, opened 2021-05-22
net: Add hooks for HTTP-request header processing

cjpatton:
This defines a new interface, net/http.CFRequestProcessor, that is invoked while processing the client's request header from the wire. It allows us to record information about the request  header that is otherwise not exposed by net/http.Request, e.g., the order of the headers.


cjpatton:
Closing in favor of #92.

PR #84, opened 2021-05-26
Fix Gopher image link in Readme file

KushagraMehta:


cjpatton:
We're going to hold off on this until we rebase on upstream Go. Thanks for sending! 

For future reference, please open an issue rather than send a PR: https://github.com/cloudflare/go/wiki/Contributing

PR #89, opened 2021-06-24
crypto/tls: Upgrade draft-ietf-tls-esni-10 to 11

cjpatton:
Drops support for the previous version of ECH and add support for the
new one. There's one caveat, however. In draft-ietf-tls-esni-11, the
ECHConfig.version uses the same codepoint as in the previous draft. This
is a bug in draft 11 that will be fixed in a future draft. In the
meantime, we use the codepoint from draft 11.

### DIFF
--- a/src/crypto/tls/ech.go
+++ b/src/crypto/tls/ech.go
@@ -14,20 +14,25 @@ import (
 
 const (
 	// Constants for TLS operations
-	echAcceptConfirmationLabel = "ech accept confirmation"
+	echAcceptConfLabel    = "ech accept confirmation"
+	echAcceptConfHRRLabel = "hrr ech accept confirmation"
 
 	// Constants for HPKE operations
 	echHpkeInfoSetup = "tls ech"
+
+	// When sent in the ClientHello, the first byte of the payload of the ECH
+	// extension indicates whether the message is the ClientHelloOuter or
+	// ClientHelloInner.
+	echClientHelloOuterVariant uint8 = 0
+	echClientHelloInnerVariant uint8 = 1
+)
+
+var (
+	zeros = [8]byte{}
 )
 
 // TODO(cjpatton): "[When offering ECH, the client] MUST NOT offer to resume any
 // session for TLS 1.2 and below [in ClientHelloInner]."
-//
-// TODO(cjpatton): "[When offering ECH, the client] MUST NOT include the
-// "pre_shared_key" extension [in ClientHelloOuter]." (This is a "don't stick
-// out" issue.)
-//
-// TODO(cjpatton): Implement client-side padding.
 func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *clientHelloMsg, err error) {
 	config := c.config
 
@@ -51,7 +56,6 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
 
 		// Grease ECH.
 		helloBase.ech = c.ech.dummy
-		helloBase.echIsOuter = true
 		c.ech.offered = false
 		c.ech.greased = true
 		helloBase.raw = nil
@@ -86,7 +90,7 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
 	helloInner = new(clientHelloMsg)
 	*helloInner = *helloBase
 	helloInner.random = c.ech.innerRandom
-	helloInner.echIsInner = true
+	helloInner.ech = []byte{echClientHelloInnerVariant}
 
 	// Ensure that only TLS 1.3 and above are offered.
 	if v := helloInner.supportedVersions; len(v) == 0 || v[len(v)-1] < VersionTLS13 {
@@ -133,22 +137,23 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
 	hello.random = helloBase.random
 	hello.sessionId = helloBase.sessionId
 	hello.serverName = hostnameInSNI(c.ech.publicName)
-	hello.echIsOuter = true
-	if testingECHOuterIsInner {
-		hello.echIsInner = true
-	}
 
 	// ClientECH, the payload of the "encrypted_client_hello" extension.
-	var ech echClient
-	_, kdfId, aeadId := c.ech.sealer.Suite().Params()
-	ech.handle.suite.kdfId = uint16(kdfId)
-	ech.handle.suite.aeadId = uint16(aeadId)
+	var ech echClientOuter
+	_, kdf, aead := c.ech.sealer.Suite().Params()
+	ech.handle.suite.kdfId = uint16(kdf)
+	ech.handle.suite.aeadId = uint16(aead)
 	ech.handle.configId = echConfig.configId
 	ech.handle.enc = enc // Empty after HRR
-	helloOuterAad := echEncodeClientHelloOuterAAD(hello.marshal(), ech.handle.marshal())
+
+	// ClientHelloOuterAAD
+	hello.ech = ech.marshal()
+	helloOuterAad := echEncodeClientHelloOuterAAD(hello.marshal(),
+		aead.CipherLen(uint(len(encodedHelloInner))))
 	if helloOuterAad == nil {
 		return nil, nil, errors.New("tls: ech: encoding of ClientHelloOuterAAD failed")
 	}
+
 	ech.payload, err = c.ech.sealer.Seal(encodedHelloInner, helloOuterAad)
 	if err != nil {
 		return nil, nil, fmt.Errorf("tls: ech: seal failed: %s", err)
@@ -156,6 +161,7 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
 	if testingECHTriggerPayloadDecryptError {
 		ech.payload[0] ^= 0xff // Inauthentic ciphertext
 	}
+	ech.raw = nil
 	hello.ech = ech.marshal()
 
 	// Offer ECH.
@@ -174,20 +180,22 @@ func (c *Conn) echAcceptOrReject(hello *clientHelloMsg) (*clientHelloMsg, error)
 		return hello, nil
 	}
 
-	if hello.echIsInner {
-		if hello.echIsOuter {
-			c.sendAlert(alertIllegalParameter)
-			return hello, errors.New("ech: hello marked as inner and outer")
-		}
-		if len(hello.ech) > 0 {
-			c.sendAlert(alertIllegalParameter)
-			return hello, errors.New("ech_is_inner: got non-empty payload")
-		}
-		// Bypass ECH and continue as backend server.
-		return hello, nil
-	}
+	if len(hello.ech) > 0 { // The ECH extension is present
+		switch hello.ech[0] {
+		case echClientHelloInnerVariant: // inner handshake
+			if len(hello.ech) > 1 {
+				c.sendAlert(alertIllegalParameter)
+				return nil, errors.New("ech: inner handshake has non-empty payload")
+			}
 
-	if !hello.echIsOuter {
+			// Bypass ECH and continue as backend server.
+			return hello, nil
+		case echClientHelloOuterVariant: // outer handshake
+		default:
+			c.sendAlert(alertIllegalParameter)
+			return nil, errors.New("ech: inner handshake has non-empty payload")
+		}
+	} else {
 		if c.ech.offered {
 			// This occurs if the server accepted prior to HRR, but the client
 			// failed to send the ECH extension in the second ClientHelloOuter. This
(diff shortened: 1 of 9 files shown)


src/crypto/tls/ech.go
```diff
@@ -327,29 +326,37 @@ func (c *Conn) echAcceptOrReject(hello *clientHelloMsg) (*clientHelloMsg, error)
...
+	s := cryptobyte.String(raw)
+	ech := new(echClientOuter)
 	ech.raw = raw
 
+	// Make sure this is the outer handshake.
+	var variant uint8
+	if !s.ReadUint8(&variant) {
+		return nil, fmt.Errorf("error parsing ClientECH.type")
+	}
+	if variant != 0 {
```
chris-wood: We seem to be using this enum a lot -- can we make it a constant so the code's more readable? E.g.,

```suggestion
	if variant != outerClientHelloVariant {
```

We could also then drop comments like the one on L376:

```
b.AddUint8(0) // Indicates ClientHelloOuter.
```

(This isn't blocking, of course. Feel free to resolve if there's no interest in this!)

src/crypto/tls/ech.go
```diff
@@ -703,9 +707,8 @@ func echDecodeClientHelloInner(encodedData, outerData, outerSessionId []byte) []
 }
 
 // echEncodeClientHelloOuterAAD interprets outerData as ClientHelloOuter and
-// handleData as an ECH context handle and maps these to a ClientHelloOuterAAD.
-// Returns nil if parsing outerData fails.
-func echEncodeClientHelloOuterAAD(outerData, handleData []byte) []byte {
+// constructs a ClientHelloOuterAAD.
```
chris-wood:
```suggestion
// constructs a ClientHelloOuterAAD. The output does not include a length prefix wrapper.
```

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -245,6 +284,45 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
...
+			echAcceptConfHRRTranscript := cloneHash(hs.transcriptInner, hs.suite.hash)
+			if echAcceptConfHRRTranscript == nil {
+				c.sendAlert(alertInternalError)
+				return errors.New("tls: internal error: failed to clone hash")
+			}
+
+			signal := hs.serverHello.ech
+			hs.serverHello.ech = []byte{0, 0, 0, 0, 0, 0, 0, 0}
+			hs.serverHello.raw = nil
+			echAcceptConfHRRTranscript.Write(hs.serverHello.marshal())
```
chris-wood: Will this marshal call ensure that the order in which extensions was received from the server is maintained? (Surely it must since the tests are presumably passing, but I thought I'd ask just in case.)

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -245,6 +284,45 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
...
+			echAcceptConfHRR := hs.suite.expandLabel(
+				hs.suite.extract(hs.helloInner.random, nil),
+				echAcceptConfHRRLabel,
+				echAcceptConfHRRTranscript.Sum(nil),
+				8)
+
+			// If this is a test, then assert the client and server have
+			// computed the acceptance signal correctly.
+			//
+			// TODO(cjpatton): Remove this assertion and use the acceptance
```
chris-wood, line 307: Is this comment up to date? (It seems to reference some old code before -11)

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -684,25 +706,29 @@ func illegalClientHelloChange(ch, ch1 *clientHelloMsg) bool {
...
-	hs.handshakeSecret = hs.suite.extract(hs.sharedKey,
-		hs.suite.deriveSecret(earlySecret, "derived", nil))
-
 	// Confirm ECH acceptance.
-	if hs.clientHello.echIsInner {
-		confTranscript := cloneHash(hs.transcript, hs.suite.hash)
-		if confTranscript == nil {
+	if hs.echIsInner() {
+		// Clear the last 8 bytes of the ServerHello.random in preparation for
+		// computing the confirmation hint.
```
chris-wood, line 708:
```suggestion
		// computing the confirmation signal.
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -718,6 +744,13 @@ func (hs *serverHandshakeStateTLS13) sendServerParameters() error {
 		return err
 	}
 
+	earlySecret := hs.earlySecret
```
chris-wood, line 743: Out of curiosity, why move this code after the acceptance signal's computed?

chris-wood (commented):
This looks good! I went through the change log and checked that each of the changes are accounted for. I've only a couple questions:

- Do we support CHInner resumption yet? I was looking for code that added a GREASEd outer PSK if one was present in CHInner (#414), but couldn't find it. (Mirroring the "early_data" extension a la #415 is related.)
- Where in the echAcceptOrReject path do we validate that the decrypted and unmarshalled CHInner has versions 1.3 or above (and otherwise aborts accordingly), enforcing #411?
- Validating ECHConfig.contents.public_name (#413, #456) is still a TODO, right?

src/crypto/tls/ech.go
cjpatton: Done (good idea!).

src/crypto/tls/ech.go
cjpatton: Done.

src/crypto/tls/handshake_client_tls13.go
cjpatton: Good catch! Added a new function, echEncodeAcceptConfHelloRetryRequest(), that encodes this properly.

No, this isn't guaranteed, and the tests don't catch this. I've been meaning to think about how design tests for this but haven't gotten around to it.

src/crypto/tls/handshake_client_tls13.go
cjpatton: This comment is up to do date. We won't refactor the client state machine until #91.

src/crypto/tls/handshake_server_tls13.go
cjpatton: I removed the comment because it's not actually relevant here. (It was copy-pasted from the SH confirmation signal computation.)

src/crypto/tls/handshake_server_tls13.go
cjpatton: This change actually moves this chunk of code back to where it was before we implemented draft-09. (See `git diff go1.1.6 -- handshake_server_tls13.go`.)

cjpatton (commented):
>     * Do we support CHInner resumption yet? I was looking for code that added a GREASEd outer PSK if one was present in CHInner (#414), but couldn't find it. (Mirroring the "early_data" extension a la #415 is related.)

Not yet. I've added TODOs where appropriate.

>     * Where in the echAcceptOrReject path do we validate that the decrypted and unmarshalled CHInner has versions 1.3 or above (and otherwise aborts accordingly), enforcing #411?

Good catch, this was not enforced. I added these checks to echAcceptOrReject().

>     * Validating ECHConfig.contents.public_name (#413, #456) is still a TODO, right?

Yes, I've added TODOs where appropriate.

cjpatton:
Rebased.

chris-wood (approved):
Thanks -- this LGTM after the new changes! 👍

cjpatton:
Rebased.

src/crypto/tls/ech.go
```diff
@@ -133,29 +133,31 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
...
+	ech.handle.suite.kdfId = uint16(kdf)
+	ech.handle.suite.aeadId = uint16(aead)
 	ech.handle.configId = echConfig.configId
 	ech.handle.enc = enc // Empty after HRR
-	helloOuterAad := echEncodeClientHelloOuterAAD(hello.marshal(), ech.handle.marshal())
+
+	// ClientHelloOuterAAD
+	hello.ech = ech.marshal()
+	helloOuterAad := echEncodeClientHelloOuterAAD(hello.marshal(),
+		int(aead.CipherLen(uint(len(encodedHelloInner)))))
```
claucece: This is a lot of converting.. can it be made better?

src/crypto/tls/ech.go
```diff
@@ -321,35 +326,56 @@ func (c *Conn) echAcceptOrReject(hello *clientHelloMsg) (*clientHelloMsg, error)
...
 
+	// Check for a well-formed ECH extension.
+	if len(helloInner.ech) != 1 ||
+		helloInner.ech[0] != echClientHelloInnerVariant {
+		c.sendAlert(alertIllegalParameter)
+		return nil, fmt.Errorf("ech: ClientHelloInner does not have a well-formed ECH extension")
+	}
+
+	// Check that the client did not offer TLS 1.2 or below.
+	if v := helloInner.supportedVersions; len(v) == 0 || v[len(v)-1] < VersionTLS13 {
```
claucece: Is this only checking only the `len(v)-1` version?

src/crypto/tls/ech.go
```diff
@@ -321,35 +326,56 @@ func (c *Conn) echAcceptOrReject(hello *clientHelloMsg) (*clientHelloMsg, error)
...
+		return nil, fmt.Errorf("unexpected ClientECH.type (want outer (0))")
+	}
+
 	// Parse the context handle.
-	s := cryptobyte.String(raw)
 	if !echReadContextHandle(&s, &ech.handle) {
 		return nil, fmt.Errorf("error parsing context handle")
 	}
-	ech.handle.raw = raw[:len(raw)-len(s)]
+	ech.handle.raw = raw[1 : len(raw)-len(s)]
```
claucece: maybe move the `len(raw)-len(s)` for their own variable for clarity?

src/crypto/tls/ech.go
```diff
@@ -643,8 +670,8 @@ func echDecodeClientHelloInner(encodedData, outerData, outerSessionId []byte) []
 					}
 					handledOuterExtensions = true
 
-					// Read the set of outer extension code points.
-					outer := make(map[uint16]bool)
+					// Read the referenced outer extensions.
+					referencedExts := make([]uint16, 0, 10)
```
claucece, line 687: The max will always be `10`?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -120,6 +125,40 @@ func (hs *clientHandshakeStateTLS13) handshake() error {
...
+	if c.ech.offered {
+		echAcceptConfTranscript := cloneHash(hs.transcriptInner, hs.suite.hash)
+		if echAcceptConfTranscript == nil {
+			c.sendAlert(alertInternalError)
+			return errors.New("tls: internal error: failed to clone hash")
+		}
+
+		sh := hs.serverHello.marshal()
+		echAcceptConfTranscript.Write(sh[:30])
+		echAcceptConfTranscript.Write([]byte{0, 0, 0, 0, 0, 0, 0, 0})
```
claucece: I think you can set the length and assing `0` just once...

claucece (commented):
Looking good!

src/crypto/tls/ech.go
cjpatton: Ha, yeah, a bit. I shaved off the outer `int()` by changing the type of the second parameter passed to `echEncodeClientHelloOuterAAD` to `uint`.

src/crypto/tls/ech.go
cjpatton: Good catch! The intention of this check is to make sure that the inner CH we constructed before with `makeClientHello()` only advertises 1.3 and above. What we actually check is that the last version in the `supported_versions` extension of the CH is less than 1.3.

These would only be equivalent if the client was required to send versions in order from highest supported version to lowest _and_ the server enforced this. Neither of these is true, AFAICT, so I fixed the check to scan the entire list.

src/crypto/tls/ech.go
cjpatton: Done.

src/crypto/tls/ech.go
cjpatton: No, it's possible for there to be more, but we don't know how many at this point. Allocating enough space for `10` initially seemed like a a sensible prediction. If we go over, the runtime will just allocate more memory as needed.

src/crypto/tls/handshake_client_tls13.go
cjpatton: how does this work, syntactically?

cjpatton: Fixed this by declaring a global `var zeros = [8]byte{}` and referring to it here and elsewhere.

src/crypto/tls/ech.go
claucece: perfect!

claucece (approved):

cjpatton:
Squashed.

PR #91, opened 2021-06-25
crypto/tls: Refactor client state machine after draft-ietf-tls-esni upgrade

cjpatton:
Based on #89.

Before draft-ietf-tls-esni-11, there was no way for a client that
offered ECH to tell whether a HelloRetryRequest (HRR) was sent by the
client-facing or backend server. Starting in draft-ietf-tls-esni-11
there is an explicit signal of ECH acceptance after HRR. This allows us
to simplify the client state machine.

### DIFF
--- a/src/crypto/tls/ech.go
+++ b/src/crypto/tls/ech.go
@@ -31,138 +31,77 @@ var (
 	zeros = [8]byte{}
 )
 
+// echOfferOrGrease is called by the client after generating its ClientHello
+// message to decide if it will offer or GREASE ECH. It does neither if ECH is
+// disabled. Returns a pair of ClientHello messages, hello and helloInner. If
+// offering ECH, these are the ClienthelloOuter and ClientHelloInner
+// respectively. Otherwise, hello is the ClientHello and helloInner == nil.
+//
 // TODO(cjpatton): "[When offering ECH, the client] MUST NOT offer to resume any
 // session for TLS 1.2 and below [in ClientHelloInner]."
 func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *clientHelloMsg, err error) {
 	config := c.config
 
-	if !config.ECHEnabled ||
-		(c.hrrTriggered && testingECHTriggerBypassAfterHRR) ||
-		(!c.hrrTriggered && testingECHTriggerBypassBeforeHRR) {
+	if !config.ECHEnabled || testingECHTriggerBypassBeforeHRR {
 		// Bypass ECH.
 		return helloBase, nil, nil
 	}
 
+	// Choose the ECHConfig to use for this connection. If none is available, or
+	// if we're not offering TLS 1.3 or above, then GREASE.
 	echConfig := config.echSelectConfig()
 	if echConfig == nil || config.maxSupportedVersion() < VersionTLS13 {
-		// Compute artifacts that are reused across HRR.
-		if c.ech.dummy == nil {
-			// Serialized ClientECH.
-			c.ech.dummy, err = echGenerateDummyExt(config.rand())
-			if err != nil {
-				return nil, nil, fmt.Errorf("tls: ech: failed to generate grease ECH: %s", err)
-			}
+		var err error
+
+		// Generate a dummy ClientECH.
+		helloBase.ech, err = echGenerateGreaseExt(config.rand())
+		if err != nil {
+			return nil, nil, fmt.Errorf("tls: ech: failed to generate grease ECH: %s", err)
 		}
 
-		// Grease ECH.
-		helloBase.ech = c.ech.dummy
+		// GREASE ECH.
 		c.ech.offered = false
 		c.ech.greased = true
 		helloBase.raw = nil
 		return helloBase, nil, nil
 	}
 
-	// Compute artifacts that are reused across HRR.
+	// Store the config id in case of HelloRetryRequest (HRR).
+	c.ech.configId = echConfig.configId
+
+	// Generate the HPKE context. Store it in case of HRR.
 	var enc []byte
-	if c.ech.sealer == nil {
-		// HPKE context.
-		enc, c.ech.sealer, err = echConfig.setupSealer(config.rand())
-		if err != nil {
-			return nil, nil, fmt.Errorf("tls: ech: %s", err)
-		}
-
-		// ECHConfig.contents.public_name.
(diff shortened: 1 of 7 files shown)


cjpatton:
Rebased.

src/crypto/tls/ech.go
```diff
@@ -27,164 +27,160 @@ const (
...
-	// "ech_is_inner" extension indicating that this is the ClientHelloInner.
-	helloInner = new(clientHelloMsg)
-	*helloInner = *helloBase
-	helloInner.random = c.ech.innerRandom
-	helloInner.ech = []byte{echClientHelloInnerVariant}
+	// ClientHelloInner is constructed from the base ClientHello. The payload of
+	// the "encrypted_client_hello" extension is a single 1 byte indicating that
+	// this is the ClientHelloInner.
+	helloInner = helloBase
+	helloInner.ech = []byte{1}
```
chris-wood:
```suggestion
	helloInner.ech = []byte{echClientHelloInnerVariant}
```

src/crypto/tls/ech.go
```diff
@@ -27,164 +27,160 @@ const (
...
 
-func (c *Conn) echAcceptOrReject(hello *clientHelloMsg) (*clientHelloMsg, error) {
+// echAcceptOrReject is called by the client-facing server to determine whether
+// ECH was offered by the client, and if so, whether to accept or reject. The
+// return value is the ClientHello that will be used for the connection.
+//
+// This function is called prior to processing the ClientHello. In case of
+// HelloRetryRequest, it is also called before processing the second
+// ClientHello. This is indicated by the afterHRR flag.
+func (c *Conn) echAcceptOrReject(hello *clientHelloMsg, afterHRR bool) (*clientHelloMsg, error) {
```
chris-wood, line 175:
```suggestion
func (c *Conn) echAcceptOrRejectAfterHRR(hello *clientHelloMsg, afterHRR bool) (*clientHelloMsg, error) {
```
(To avoid the comments when invoking this function, i.e., `echAcceptOrRejectAfterHRR(..., false /* afterHRR */)`

chris-wood (commented):
As I understand it, the main (non-cosmetic) changes are:

1. Unify the ECH and non-ECH paths behind `echOfferOrGrease`, and
2. Simplify the HRR detection path by checking the signal

If there are other significant changes, they didn't stand out to me. @cjpatton, can you clarify whether or not I missed something?

src/crypto/tls/ech.go
cjpatton: Done.

src/crypto/tls/ech.go
cjpatton: This function is called before and after HRR (hence the `afterHRR` flag).

chris-wood: Yeah, I get the purpose of the flag, it just seems like `echAcceptOrRejectAfterHRR` is a more descriptive name. If I see this:

```
echAcceptOrRejectAfterHRR(hello, false)
```

I can reasonably assume that I'm processing `hello` before HRR. But if I see this:

```
echAcceptOrReject(hello, false)
```

It's less clear what `false` means, which is why the comments `/* afterHRR */` are included. 🤷

cjpatton:
> If there are other significant changes, they didn't stand out to me. @cjpatton, can you clarify whether or not I missed something?

No other significant changes. One minor change: I moved the logic for selecting the OuterExtensions out of echOfferorGrease().

src/crypto/tls/ech.go
cjpatton: The convention in this library is to pass a flag and not try to name the function in a way that make it clear in context what the flag does. I'll stick with that convention.

chris-wood (approved):

src/crypto/tls/ech.go
```diff
@@ -1008,3 +1009,50 @@ func (c *Config) echCanAccept() bool {
...
+	// EncodedClientHelloInner is constructed from ClientHelloInner by removing
+	// extensions that are also used in the ClientHelloOuter.
+	//
+	// NOTE(cjpatton): It would be nice to incorporate more extensions, but
+	// "key_share" is the last extension to appear in the ClientHello before
+	// "pre_shared_key". As a result, the only contiguous sequence of outer
+	// extensions that contains "key_share" is "key_share" itself. Note that
+	// we cannot change the order of extensions in the ClientHello, as the
+	// unit tests expect "key_share" to be the second to last extension.
+	echOuterExtensions = []uint16{extensionKeyShare}
```
wbl: Why is this stuff here and not in test_ech.go?

wbl (approved):

src/crypto/tls/ech.go
cjpatton: Hmmm, well I suppose we could have two `init()` functions, one here and one in test_ech.go, where the latter overwrites the value of `echOuterExtensions` set here. But I think it's more readable to have the logic for setting this value all in one function.

cjpatton:
Rebased.

src/crypto/tls/ech.go
claucece: `init` is rarely used in the library outside tests. Does this need to be an `init`?

src/crypto/tls/handshake_server.go
```diff
@@ -142,7 +142,7 @@ func (c *Conn) readClientHello() (*clientHelloMsg, error) {
 	// or GetCertifciate(). Hence, it is not currently possible to reject ECH if
 	// we don't recognize the inner SNI. This may or may not be desirable in the
 	// future.
-	clientHello, err = c.echAcceptOrReject(clientHello)
+	clientHello, err = c.echAcceptOrReject(clientHello, false /* afterHRR */)
```
claucece: I think is better to move the comment outside the function.

claucece (changes requested):
Thanks for this!

cjpatton:
Rebased and squashed.

src/crypto/tls/ech.go
cjpatton: Nope! I replaced the `init()` with a function `echOuterExtensions()` that generates this list.

src/crypto/tls/handshake_server.go
cjpatton: Done (here and in handshake_server_tls13.go).

claucece (approved):
LGTM ;) Thanks!

cjpatton:
Squashed.

PR #92, opened 2021-06-29
net/http: Preserve some features of headers on the wire

cjpatton:
Adds a field to net/http.Request that allows us to reconstruct the HTTP headers (almost) as they appeared on the wire.

### DIFF
--- a/src/net/http/request.go
+++ b/src/net/http/request.go
@@ -315,6 +315,14 @@ type Request struct {
 	// redirects.
 	Response *Response
 
+	// CFHeaderLines represents the headers in a way that allows them to be
+	// re-constructed (almost) as they appeared on the wire.
+	//
+	// NOTE(cjpatton): The "CF" prefix denotes the fact that this parameter is
+	// used for a Cloudflare-internal purpose. It may or may not be useful for
+	// upstream Go.
+	CFHeaderLines []textproto.CFHeaderLine
+
 	// ctx is either the client or server context. It should only
 	// be modified via copying the whole Request using WithContext.
 	// It is unexported to prevent people from using Context wrong
@@ -1009,16 +1017,27 @@ func putTextprotoReader(r *textproto.Reader) {
 // requests and handle them via the Handler interface. ReadRequest
 // only supports HTTP/1.x requests. For HTTP/2, use golang.org/x/net/http2.
 func ReadRequest(b *bufio.Reader) (*Request, error) {
-	return readRequest(b, deleteHostHeader)
+	return readRequest(b, deleteHostHeader, cfRecordRequestLines)
 }
 
-// Constants for readRequest's deleteHostHeader parameter.
+// CFReadRequest is like ReadRequest except that it sets CFHeaderLines in the
+// request.
+//
+// NOTE(cjpatton): The "CF" prefix denotes the fact that this parameter is used
+// for a Cloudflare-internal purpose. It may or may not be useful for upstream
+// Go.
+func CFReadRequest(b *bufio.Reader) (*Request, error) {
+	return readRequest(b, deleteHostHeader, true)
+}
+
+// Constants for readRequest's flags.
 const (
-	deleteHostHeader = true
-	keepHostHeader   = false
+	deleteHostHeader     = true
+	keepHostHeader       = false
+	cfRecordRequestLines = false
 )
 
-func readRequest(b *bufio.Reader, deleteHostHeader bool) (req *Request, err error) {
+func readRequest(b *bufio.Reader, deleteHostHeader, cfRecordRequestLines bool) (req *Request, err error) {
 	tp := newTextprotoReader(b)
 	req = new(Request)
 
@@ -1071,11 +1090,12 @@ func readRequest(b *bufio.Reader, deleteHostHeader bool) (req *Request, err erro
 	}
 
 	// Subsequent lines: Key: value.
-	mimeHeader, err := tp.ReadMIMEHeader()
+	mimeHeader, orderedHeader, err := tp.CFReadMIMEHeader(cfRecordRequestLines)
 	if err != nil {
 		return nil, err
 	}
 	req.Header = Header(mimeHeader)
+	req.CFHeaderLines = orderedHeader
 
 	// RFC 7230, section 5.3: Must treat
 	//	GET /index.html HTTP/1.1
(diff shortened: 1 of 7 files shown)


src/net/textproto/reader.go
```diff
@@ -490,28 +495,31 @@ func (r *Reader) ReadMIMEHeader() (MIMEHeader, error) {
 		strs = make([]string, hint)
 	}
 
+	ordered := make([]CFHeaderLine, 0, hint)
```
Lekensteyn: Can you make all of this conditional via a flag in `http.Server`? If this information is not needed, then these allocations can be skipped. This also avoids the need to set values to `nil` during `request_test.go` as well as all those other tests.

Lekensteyn (commented):
Looks fine for the most part except for enabling the collection by default which affects other tests.

src/net/textproto/reader.go
cjpatton: Done.

src/net/http/request.go
```diff
@@ -1071,11 +1080,14 @@ func readRequest(b *bufio.Reader, deleteHostHeader bool) (req *Request, err erro
 	}
 
 	// Subsequent lines: Key: value.
-	mimeHeader, err := tp.ReadMIMEHeader()
+	mimeHeader, orderedHeader, err := tp.CFReadMIMEHeader()
 	if err != nil {
 		return nil, err
 	}
 	req.Header = Header(mimeHeader)
+	if cfRecordRequestLines {
```
Lekensteyn: The global `cfRecordRequestLines` variable has to be removed and it needs to somehow use the `net/http.Server` field.

Also just set `CFHeaderLines` unconditionally here `CFReadMIMEHeader` *should* return `nil` when the feature is not enabled.

src/net/textproto/reader.go
```diff
@@ -481,6 +481,15 @@ func (r *Reader) ReadDotLines() ([]string, error) {
 //	}
 //
 func (r *Reader) ReadMIMEHeader() (MIMEHeader, error) {
+	header, _, err := r.readMIMEHeader(false /* cfRecordRequestLines */)
+	return header, err
+}
+
+func (r *Reader) CFReadMIMEHeader() (MIMEHeader, []CFHeaderLine, error) {
+	return r.readMIMEHeader(true /* cfRecordRequestLInes */)
```
Lekensteyn: `I` in `LInes is accidentally capitalized. However I believe that you have to expose this bool parameter anyway (and set it to the http.Server parameter).

Lekensteyn (changes requested):

src/net/http/request.go
cjpatton: `cfRecordRequestLines` isn't a global variable here, it's a parameter of function `readRequest`. The parameter *is* set to a global variable when called by `ReadRequest`, but it's set by `Server.CFRecordRequestLines` when called by the server.

src/net/textproto/reader.go
cjpatton: Added `recordHeaderLInes bool` parameter to `CFReadMIMEHeader`. In net/http this is set to `Server.CFRecordRecordLInes`.

wbl (dismissed):

src/net/http/request.go
Lekensteyn: > The parameter is set to a global variable when called by ReadRequest

Yes that is exactly what I mean, you introduce a new global:
```diff
 // only supports HTTP/1.x requests. For HTTP/2, use golang.org/x/net/http2.
 func ReadRequest(b *bufio.Reader) (*Request, error) {
-       return readRequest(b, deleteHostHeader)
+       return readRequest(b, deleteHostHeader, cfRecordRequestLines)
 }

-// Constants for readRequest's deleteHostHeader parameter.
+// Constants for readRequest's flags.
 const (
        deleteHostHeader     = true
        keepHostHeader       = false
+       cfRecordRequestLines = false
 )
```

However after looking at it again, its sole purpose appears to be documentation. In `server.go`, `conn.readRequest` does propagate the `Server.CFRecordRequestLines` flag:

```diff
-       req, err := readRequest(c.bufr, keepHostHeader)
+       req, err := readRequest(c.bufr, keepHostHeader, c.server.CFRecordRequestLines)
        if err != nil {
```

Consider this concern resolved.

src/net/textproto/reader.go
```diff
@@ -481,6 +481,15 @@ func (r *Reader) ReadDotLines() ([]string, error) {
 //	}
 //
 func (r *Reader) ReadMIMEHeader() (MIMEHeader, error) {
+	header, _, err := r.readMIMEHeader(false /* recordRequestLines */)
+	return header, err
+}
+
+func (r *Reader) CFReadMIMEHeader(recordRequestLines bool) (MIMEHeader, []CFHeaderLine, error) {
+	return r.readMIMEHeader(recordRequestLines)
```
Lekensteyn: This function can be removed now and the other function below can be renamed to `CFReadMIMEHeader` right?

Lekensteyn (dismissed):
LGTM, one minor nit

src/net/textproto/reader.go
cjpatton: Done.

Lekensteyn (approved):

wbl (approved):

src/net/http/request.go
```diff
@@ -1009,16 +1017,27 @@ func putTextprotoReader(r *textproto.Reader) {
 // requests and handle them via the Handler interface. ReadRequest
 // only supports HTTP/1.x requests. For HTTP/2, use golang.org/x/net/http2.
 func ReadRequest(b *bufio.Reader) (*Request, error) {
-	return readRequest(b, deleteHostHeader)
+	return readRequest(b, deleteHostHeader, cfRecordRequestLines)
+}
+
+// CFReadREquest is like ReadRequest except that it sets CFHeaderLines in the
```
Lekensteyn:
```suggestion
// CFReadRequest is like ReadRequest except that it sets CFHeaderLines in the
```

Lekensteyn (approved):
small spelling nit. Other than that exposing this flag for simplification of tests sounds reasonable. The alternative (tests requiring sending manual TCP payloads and a special HTTP Handler) is also possible but slightly more complicated.

PR #95, opened 2021-07-13
crypto/tls: Reject ECH on invalid encapsulated key

cjpatton:
The ECH provider aborts if it gets an encapsulated key it's unable to parse, causing the ECH server to abort with an "internal_error" alert. This condition can be triggered by a client that generates a GREASE ECH extension with a config_id that happens to match a known config, but uses the incorrect KEM algorithm. This changes the server behavior so that it rejects instead.

The key text from the spec that tells us what to do here is in [Section 7.1 (Client-facing Server Behavior)](https://www.ietf.org/archive/id/draft-ietf-tls-esni-12.html#name-server-behavior): "If decryption fails, the server continues to the next candidate ECHConfig." Being unable to parse the KEM encapsulated key certainly counts as a decryption failure; and because we have exhausted our list of ECH configs, we must reject.

### DIFF
--- a/src/crypto/tls/ech_provider.go
+++ b/src/crypto/tls/ech_provider.go
@@ -47,6 +47,8 @@ const (
 	ECHProviderSuccess ECHProviderStatus = 0
 	ECHProviderReject                    = 1
 	ECHProviderAbort                     = 2
+
+	errHPKEInvalidPublicKey = "hpke: invalid KEM public key"
 )
 
 // ECHProviderResult represents the result of invoking the ECH provider.
@@ -179,6 +181,17 @@ func (keySet *EXP_ECHKeySet) GetDecryptionContext(rawHandle []byte, version uint
 	// Compute the decryption context.
 	opener, err := key.setupOpener(handle.enc, suite)
 	if err != nil {
+		if err.Error() == errHPKEInvalidPublicKey {
+			// This occurs if the KEM algorithm used to generate handle.enc is
+			// not the same as the KEM algorithm of the key. One way this can
+			// happen is if the client sent a GREASE ECH extension with a
+			// config_id that happens to match a known config, but which uses a
+			// different KEM algorithm.
+			res.Status = ECHProviderReject
+			res.RetryConfigs = keySet.configs
+			return // Reject
+		}
+
 		res.Status = ECHProviderAbort
 		res.Alert = uint8(alertInternalError)
 		res.Error = err
--- a/src/crypto/tls/tls_cf_test.go
+++ b/src/crypto/tls/tls_cf_test.go
@@ -1,6 +1,8 @@
 package tls
 
 import (
+	"circl/hpke"
+	"crypto/rand"
 	"testing"
 )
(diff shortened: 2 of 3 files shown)


chris-wood (approved):

Lekensteyn (approved):
The fix looks reasonable, could we have a regression test for this just in case we upgrade the circl library and it changed its errors?

src/crypto/tls/tls_cf_test.go
```diff
@@ -17,3 +19,49 @@ func TestPropagateCFControl(t *testing.T) {
 		t.Errorf("failed to propagate CFControl: got %v; want %v", got, want)
 	}
 }
+
+// If the client uses the wrong KEM algorithm to offer ECH, the ECH provider
+// should reject rather than abort. We check for this condition by looking at
+// the error returned by hpke.Reciver.Setup(). This test asserts that the
```
Lekensteyn:
```suggestion
// the error returned by hpke.Receiver.Setup(). This test asserts that the
```

Lekensteyn: And perhaps the comment could explicitly point to the `EXP_ECHKeySet.GetDecryptionContext`?

Alternatively, how feasible would a unit test be for that function?

Lekensteyn (approved):
thanks, one question but looks good otherwise

cjpatton:
TODO(@cjpatton) Replace the CIRCL regression test with an ECH unit test for the following case: he client uses an invalid ECH config with a config_id matching a valid ECH config, but with a different KEM algorithm. The client-facing server should reject. (See https://github.com/cloudflare/go/pull/95#discussion_r672450095.)

I'll take care of this once #89 and #96 land.

claucece (approved):
LGTM.

cjpatton:
Rebased.

src/crypto/tls/ech_provider.go
```diff
@@ -179,6 +179,17 @@ func (keySet *EXP_ECHKeySet) GetDecryptionContext(rawHandle []byte, version uint
 	// Compute the decryption context.
 	opener, err := key.setupOpener(handle.enc, suite)
 	if err != nil {
+		if err.Error() == "hpke: invalid KEM public key" {
```
claucece: Mmm.. can we make the string a variable to be compared?

claucece (commented):
Looking nice!

cjpatton:
Rebased and squashed.

cjpatton:
Rebased.

src/crypto/tls/ech_provider.go
cjpatton: Done.

Lekensteyn (approved):
LGTM, don't forget to squash and edit the commit message

cjpatton:
Rebased and squashed, waiting on CI.

cjpatton:
Rebased after merging https://github.com/cloudflare/go/pull/96, letting CI run one more time.

PR #96, opened 2021-07-13
crypto/tls: Upgrade draft-ietf-tls-esni-11 to 12

cjpatton:
Drops support for the previous draft and adds support for the current
one. This change also includes the ClientHelloInner padding scheme
described in Section 6.1.3.

### DIFF
--- a/src/crypto/tls/ech.go
+++ b/src/crypto/tls/ech.go
@@ -66,8 +66,9 @@ func (c *Conn) echOfferOrGrease(helloBase *clientHelloMsg) (hello, helloInner *c
 		return helloBase, nil, nil
 	}
 
-	// Store the config id in case of HelloRetryRequest (HRR).
+	// Store the ECH config parameters that are needed later.
 	c.ech.configId = echConfig.configId
+	c.ech.maxNameLen = int(echConfig.maxNameLen)
 
 	// Generate the HPKE context. Store it in case of HRR.
 	var enc []byte
@@ -136,7 +137,10 @@ func (c *Conn) echUpdateClientHelloOuter(hello, helloInner *clientHelloMsg, enc 
 
 	// EncodedClientHelloInner
 	helloInner.raw = nil
-	encodedHelloInner := echEncodeClientHelloInner(helloInner.marshal())
+	encodedHelloInner := echEncodeClientHelloInner(
+		helloInner.marshal(),
+		len(helloInner.serverName),
+		c.ech.maxNameLen)
 	if encodedHelloInner == nil {
 		return errors.New("tls: ech: encoding of EncodedClientHelloInner failed")
 	}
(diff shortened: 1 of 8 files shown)


chris-wood (approved):

src/crypto/tls/ech.go
```diff
@@ -1056,3 +1080,10 @@ func echCopyExtensionFromClientHelloInner(hello, helloInner *clientHelloMsg, ext
 		panic(fmt.Errorf("tried to copy unrecognized extension: %04x", ext))
 	}
 }
+
+func max(a, b int) int {
```
claucece: If this is a function for the `int` data type, shouldn't it be there?

src/crypto/tls/ech.go
```diff
@@ -584,6 +588,18 @@ func echEncodeClientHelloInner(innerData []byte) []byte {
 		panic(err) // Host encountered internal error
 	}
 
+	// Add padding.
+	paddingLen := 0
+	if serverNameLen > 0 {
+		paddingLen += max(0, maxNameLen-serverNameLen)
+	} else {
+		paddingLen += maxNameLen + 9
```
claucece: Is this '9' an arbitrary number? if it is not, it might be good to make it a constant.

src/crypto/tls/ech.go
```diff
@@ -723,6 +739,14 @@ func echDecodeClientHelloInner(encodedData, outerData, outerSessionId []byte) []
 		panic(err) // Host encountered internal error
 	}
 
+	// Read the padding.
+	for !s.Empty() {
+		var zero uint8
+		if !s.ReadUint8(&zero) || zero != 0 {
```
claucece, line 777: It might be good to change the order here to `!s.ReadUint8(&zero) || zero != 0`. I think golang executes first the left-hand and then the right-hand, so maybe you want first to check the latter rather than the first as that check might be faster...

cjpatton: The first expression (`!s.ReadUint8(&zero)`) needs to be evaluated first because it sets `zero` as a side-effect.

src/crypto/tls/ech.go
cjpatton: Do you mean a built-in function or something in the standard library? This doesn't exist as far as I know.

src/crypto/tls/ech.go
cjpatton: Done. I also added some documentation to the spec to explain why `paddingLen` is calculated this way.

cjpatton:
Rebased.

src/crypto/tls/ech.go
claucece: Yes, it does not exist. But as this is modifying the `int`.. does it have to live in `tls`?

src/crypto/tls/ech.go
claucece: ah.. interesiting.

claucece (changes requested):
Just waiting the response on the `int` comment. Thanks!

cjpatton:
Rebased and squashed.

src/crypto/tls/ech.go
cjpatton: No it doesn't need to live here :) What's the name of the built-in? I'm happy to just use that.

cjpatton:
Rebased.

src/crypto/tls/ech.go
claucece: Mmm.. see, for example, `crypto/rand/util.go` is the file for this 'util-like' functions. Maybe something like it?

cjpatton: Oh, I think I see. You're suggestion is to move this function out of `ech.go` since it's useful more generally? A good place might be `tls_cf.go` in this directory.

In any case, I decided to solve this by just getting rid of the `max()` function, since it's only used in one place.

claucece (approved):
LGTM ;)

cjpatton:
Rebased and squashed, waiting on CI.

PR #97, opened 2021-07-16
crypto/tls: Propagate ClientHello through ConnectionState

cjpatton:
Adds a field to crypto/tls.ConnectionState containing the serialized
ClientHello used to terminate the connection.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -350,6 +350,13 @@ type ConnectionState struct {
 	// This feature is unstable and applications MUST NOT depend on it.
 	CFControl interface{}
 
+	// CFClientHello is the serialized ClientHello handshake message used by the
+	// connection. It can be empty if the handshake has not concluded.
+	//
+	// NOTE: This feature is used for a Cloudflare-internal functionality. It
+	// is unstable and applications MUST NOT depend on it.
+	CFClientHello []byte
+
 	// ekm is a closure exposed via ExportKeyingMaterial.
 	ekm func(label string, context []byte, length int) ([]byte, error)
 }
--- a/src/crypto/tls/conn.go
+++ b/src/crypto/tls/conn.go
@@ -141,6 +141,9 @@ type Conn struct {
 	// Set by the client and server when an HRR message was sent in this
 	// handshake.
 	hrrTriggered bool
+
+	// The serialized ClientHello used in this connection.
+	clientHello []byte
 }
 
 // Access to net.Conn methods.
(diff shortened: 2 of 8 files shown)


src/crypto/tls/common.go
```diff
@@ -350,6 +350,13 @@ type ConnectionState struct {
 	// This feature is unstable and applications MUST NOT depend on it.
 	CFControl interface{}
 
+	// CFClientHello is the serialized ClientHello used to terminate the
+	// connection.
```
Lekensteyn: It might be worth pointing out that this is only set when the handshake has successfully been completed.
"terminate the connection" might be confusing who are not familiar with ECH. I suggest changing it, here is a proposal based on the existing documentation for "Version" and "HandshakeComplete".

```suggestion
	// CFClientHello is the serialized ClientHello handshake message used by the
	// connection. It can be empty if the handshake has not concluded.
```

Lekensteyn (approved):
For the given use case, it looks good to me. Minor nit on the documentation.

claucece:
I might be behind over this: are we allowing now changes that misaligned heavily with upstream (for me, it is completely ok)? At some point, what I understood was that the idea of this library was that it didn't diverge from upstream that much, and this seems like a kind of big change...

Lekensteyn:
> I might be behind over this: are we allowing now changes that misaligned heavily with upstream (for me, it is completely ok)? At some point, what I understood was that the idea of this library was that it didn't diverge from upstream that much, and this seems like a kind of big change...

I originally wanted this library to be as close to upstream as possible, but at some point I have relaxed my position to have more freedom in experimentation. The changes proposed here are needed because the serialized Client Hello is different with Encrypted Client Hello (ECH) support.

Without ECH, in theory one could wrap the underlying TCP connection and parse TLS records + TLS ClientHello handshake messages. And one would also have to take care of the Client Hello following a HelloRetryRequest message. That would already be complex, but at least it would enable the desired goal of obtaining the raw Client Hello.

However with ECH, that approach no longer works since the "inner Client Hello" is obtained by decrypting a special TLS extension (encrypted_client_hello). And one thus needs library support in order to obtain the decrypted inner Client Hello.

This will result in "fun" when rebasing on the latest major Go version, but that is a trade-off to be made 🙁

cjpatton:
> I might be behind over this: are we allowing now changes that misaligned heavily with upstream (for me, it is completely ok)? At some point, what I understood was that the idea of this library was that it didn't diverge from upstream that much, and this seems like a kind of big change...

We might instead just merge this into `cf-tls-1000` to keep `cf` clean.

claucece:
> We might instead just merge this into cf-tls-1000 to keep cf clean.

No hahaha.. I like this, so go for it haha ;)

claucece (approved):
LGTM ;)

PR #101, opened 2021-08-24
crypto/tls: Upgrade draft-ietf-tls-esni-12 to 13

cjpatton:
Drops support for the previous version of ECH and adds support for the
latest version. Changes include:

* An optional GREASE extension for the acceptance signal in the
  HelloRetryRequest, which is supported by the test server.

* The rules for parsing the "encrypted_client_hello" extension are
  stricter.

### DIFF
--- a/src/crypto/tls/handshake_server_tls13.go
+++ b/src/crypto/tls/handshake_server_tls13.go
@@ -568,8 +568,9 @@ func (hs *serverHandshakeStateTLS13) doHelloRetryRequest(selectedGroup CurveID) 
 		selectedGroup:     selectedGroup,
 	}
 
-	// Confirm ECH acceptance.
+	// Decide whether to send "encrypted_client_hello" extension.
 	if hs.echIsInner() {
+		// Confirm ECH acceptance if this is the inner handshake.
 		echAcceptConfHRRTranscript := cloneHash(hs.transcript, hs.suite.hash)
 		if echAcceptConfHRRTranscript == nil {
 			c.sendAlert(alertInternalError)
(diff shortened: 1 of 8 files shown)


chris-wood (approved):
Nice and clean =)

chris-wood:
@Lekensteyn can you please take a look?

Lekensteyn (approved):

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -587,6 +588,17 @@ func (hs *serverHandshakeStateTLS13) doHelloRetryRequest(selectedGroup CurveID)
 
 		helloRetryRequest.ech = echAcceptConfHRRSignal
 		helloRetryRequest.raw = nil
+	} else if c.ech.greased {
```
marwanfayed, line 591: Conditional here has no structural/data connection to 'if' on 572. I get that it's supposed to prioritize inner, and absent that then check if it's greased. But note that since one check is on 'hs' and other is on 'c', it reads as though "If apple is red, else if orange is orange".
Am purely flagging - not sure there is a readability or structural fix, or if one is needed at all.

cjpatton: This would be readability an issue, not a structural one. The first condition is (from `hs.echIsInner()`):
```go
len(hs.clientHello.ech) == 1 && hs.clientHello.ech[0] == echClientHelloInnerVariant
```
This holds if the CH is a ClientHelloInner in the ECh protocol. Under this condition the backend server is supposed to confirm acceptance. It might seem like an equivalent check would be `c.ech.accepted == true`, which would be a lot more readable; but because it is the backend server that is confirming acceptance, it has no way of knowing if the client-facing server accepted ECH (i.e., set `c.ech.accepted` to `true`). In theory, these servers are separated by a network.

Do you have any suggestions for how to improve readability? Maybe a comment would help?

Issue #104, opened 2021-09-16
crypto/tls: Add ECHOffered ConnectionState flag


chris-wood:
This lets folks determine if ECH was greased or rejected.

Lekensteyn:
This was done in PR #105.

PR #105, opened 2021-09-16
crypto/tls: Add ECHOffered bool to ConnectionState.

chris-wood:
See #104.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -338,6 +338,10 @@ type ConnectionState struct {
 	// accepted by the server.
 	ECHAccepted bool
 
+	// ECHOffered is set if the ECH extension is present in the ClientHello.
+	// This means the client has offered ECH or sent GREASE ECH.
+	ECHOffered bool
+
 	// CFControl is used to pass additional TLS configuration information to
 	// HTTP requests.
 	//
--- a/src/crypto/tls/conn.go
+++ b/src/crypto/tls/conn.go
@@ -1475,6 +1475,7 @@ func (c *Conn) connectionStateLocked() ConnectionState {
 	state.SignedCertificateTimestamps = c.scts
 	state.OCSPResponse = c.ocspResponse
 	state.ECHAccepted = c.ech.accepted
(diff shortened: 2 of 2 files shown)


src/crypto/tls/common.go
```diff
@@ -338,6 +338,9 @@ type ConnectionState struct {
 	// accepted by the server.
 	ECHAccepted bool
 
+	// ECHOffered is set if the ECH extension was offered by the client.
```
cjpatton: To match the verbage elsewhere in the code, this should say
```suggestion
	// ECHOffered is set if the ECH extension is present in the ClientHello.
	// This means the client has offered ECH or sent GREASE ECH.
```
Also, we might consider giving it a slightly less ambiguous name, like `ECHExtensionPresent` or something. Alternatively, you could add two flags, one for ECH offered and another for ECH greased.

src/crypto/tls/conn.go
```diff
@@ -1475,6 +1475,7 @@ func (c *Conn) connectionStateLocked() ConnectionState {
 	state.SignedCertificateTimestamps = c.scts
 	state.OCSPResponse = c.ocspResponse
 	state.ECHAccepted = c.ech.accepted
+	state.ECHOffered = c.ech.offered
```
cjpatton: `c.ech.offered == true` if ECH was offered by the client and rejected or accepted by the server.
`c.ech.greased == true` if the client sent GREASE ECH.

Hence, the condition you're after is `c.ech.offered || c.ech.greased`.

cjpatton (changes requested):

src/crypto/tls/conn.go
chris-wood: Fixed.

cjpatton (approved):
LGTM. Be sure to squash. Also, our convention for commit messages is to not include the period "."

chris-wood:
@cjpatton squashed and ready to go. @claucece, can you please take a look?

claucece (approved):
LGTM ;)

PR #112, opened 2022-07-06
crypto/tls: Add Kyber{512,768}X25519 group for hybrid key exchange

### LINKED ISSUE #111: Add support for hybrid key exchange algorithms with TLS
... based on the NIST announcement, that would be Kyber.

chris-wood:
Closes #111.

Starting as draft for now.

### DIFF
--- a/src/circl/kem/hybrid/hybrid.go
+++ b/src/circl/kem/hybrid/hybrid.go
@@ -46,8 +46,11 @@ var ErrUninitialized = errors.New("public or private key not initialized")
 // Returns the hybrid KEM of Kyber512 and X25519.
 func Kyber512X25519() kem.Scheme { return kyber512X }
 
+// Returns the hybrid KEM of Kyber768 and X25519.
+func Kyber768X25519() kem.Scheme { return kyber768X }
+
 // Returns the hybrid KEM of Kyber768 and X448.
-func Kyber768X448() kem.Scheme { return kyber768X }
+func Kyber768X448() kem.Scheme { return kyber768X4 }
 
 // Returns the hybrid KEM of Kyber1024 and X448.
 func Kyber1024X448() kem.Scheme { return kyber1024X }
@@ -59,6 +62,12 @@ var kyber512X kem.Scheme = &scheme{
 }
 
 var kyber768X kem.Scheme = &scheme{
+	"Kyber768-X25519",
+	kyber768.Scheme(),
+	hpke.KEM_X25519_HKDF_SHA256.Scheme(),
+}
+
+var kyber768X4 kem.Scheme = &scheme{
 	"Kyber768-X448",
 	kyber768.Scheme(),
 	hpke.KEM_X448_HKDF_SHA512.Scheme(),
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -28,6 +28,7 @@ type clientKeySharePrivate interface{}
 
 var (
 	kyber512X25519CurveID = CurveID(0xFF01)
+	kyber768X25519CurveID = CurveID(0xFF02)
 	invalidCurveID        = CurveID(0xFFFF)
 )
 
@@ -35,6 +36,8 @@ func kemSchemeKeyToCurveID(s kem.Scheme) CurveID {
 	switch s.Name() {
 	case "Kyber512-X25519":
 		return kyber512X25519CurveID
+	case "Kyber768-X25519":
+		return kyber768X25519CurveID
 	default:
 		return invalidCurveID
 	}
@@ -54,8 +57,11 @@ func clientKeySharePrivateCurveID(ks clientKeySharePrivate) CurveID {
 
 // Returns scheme by CurveID if supported by Circl
 func curveIdToCirclScheme(id CurveID) kem.Scheme {
-	if id == kyber512X25519CurveID {
+    switch id {
+    case kyber512X25519CurveID:
 		return hybrid.Kyber512X25519()
+    case kyber768X25519CurveID:
+		return hybrid.Kyber768X25519()
 	}
 	return nil
 }
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -17,6 +17,7 @@
 package tls
 
 import (
+	"fmt"
 	"io"
 
 	"circl/kem"
@@ -57,10 +58,10 @@ func clientKeySharePrivateCurveID(ks clientKeySharePrivate) CurveID {
 
 // Returns scheme by CurveID if supported by Circl
 func curveIdToCirclScheme(id CurveID) kem.Scheme {
-    switch id {
-    case kyber512X25519CurveID:
+	switch id {
+	case kyber512X25519CurveID:
 		return hybrid.Kyber512X25519()
-    case kyber768X25519CurveID:
+	case kyber768X25519CurveID:
 		return hybrid.Kyber768X25519()
 	}
 	return nil
@@ -72,11 +73,11 @@ func encapsulateForKem(scheme kem.Scheme, rnd io.Reader, ppk []byte) (
 	ct, ss []byte, internalErr bool, err error) {
 	pk, err := scheme.UnmarshalBinaryPublicKey(ppk)
 	if err != nil {
-		return nil, nil, false, err
+		return nil, nil, false, fmt.Errorf("unpack pk: %w", err)
 	}
 	seed := make([]byte, scheme.EncapsulationSeedSize())
 	if _, err := io.ReadFull(rnd, seed); err != nil {
-		return nil, nil, true, err
+		return nil, nil, true, fmt.Errorf("random: %w", err)
 	}
 	ct, ss, err = scheme.EncapsulateDeterministically(pk, seed)
 	return ct, ss, false, err
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -1,3 +1,6 @@
+// Copyright 2022 Cloudflare, Inc. All rights reserved. Use of this source code
+// is governed by a BSD-style license that can be found in the LICENSE file.
+//
 // Glue to add Circl's (post-quantum) hybrid KEMs to quic-go.
 //
 // To enable set CurvePreferences with the desired scheme as the first element:
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -1,7 +1,7 @@
 // Copyright 2022 Cloudflare, Inc. All rights reserved. Use of this source code
 // is governed by a BSD-style license that can be found in the LICENSE file.
 //
-// Glue to add Circl's (post-quantum) hybrid KEMs to quic-go.
+// Glue to add Circl's (post-quantum) hybrid KEMs.
 //
 // To enable set CurvePreferences with the desired scheme as the first element:
 //
@@ -13,8 +13,8 @@
 //
 //   config.CurvePreferences = []tls.CurveID{
 //      hybrid.Kyber512X25519().(tls.TLSScheme).TLSCurveID(),
-//      qtls.X25519,
-//      qtls.P256,
+//      tls.X25519,
+//      tls.P256,
 //   }
 
 package tls
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -73,17 +73,17 @@ func curveIdToCirclScheme(id CurveID) kem.Scheme {
 // Generate a new shared secret and encapsulates it for the packed
 // public key in ppk using randomness from rnd.
 func encapsulateForKem(scheme kem.Scheme, rnd io.Reader, ppk []byte) (
-	ct, ss []byte, internalErr bool, err error) {
+	ct, ss []byte, alert alert, err error) {
 	pk, err := scheme.UnmarshalBinaryPublicKey(ppk)
 	if err != nil {
-		return nil, nil, false, fmt.Errorf("unpack pk: %w", err)
+		return nil, nil, alertIllegalParameter, fmt.Errorf("unpack pk: %w", err)
 	}
 	seed := make([]byte, scheme.EncapsulationSeedSize())
 	if _, err := io.ReadFull(rnd, seed); err != nil {
-		return nil, nil, true, fmt.Errorf("random: %w", err)
+		return nil, nil, alertInternalError, fmt.Errorf("random: %w", err)
 	}
 	ct, ss, err = scheme.EncapsulateDeterministically(pk, seed)
-	return ct, ss, false, err
+	return ct, ss, alertIllegalParameter, err
 }
 
 // Generate a new keypair using randomness from rnd.
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -55,7 +55,7 @@ func clientKeySharePrivateCurveID(ks clientKeySharePrivate) CurveID {
 		if ret == invalidCurveID {
 			panic("cfkem: internal error: don't know CurveID for this KEM")
 		}
-        return ret
+		return ret
 	case ecdheParameters:
 		return v.CurveID()
 	default:
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -1,0 +1,88 @@
+// Glue to add Circl's (post-quantum) hybrid KEMs to quic-go.
+//
+// To enable set CurvePreferences with the desired scheme as the first element:
+//
+//   import (
+//      "github.com/cloudflare/circl/kem/tls"
+//      "github.com/cloudflare/circl/kem/hybrid"
+//
+//          [...]
+//
+//   config.CurvePreferences = []tls.CurveID{
+//      hybrid.Kyber512X25519().(tls.TLSScheme).TLSCurveID(),
+//      qtls.X25519,
+//      qtls.P256,
+//   }
+
+package tls
+
+import (
+	"io"
+
+	"circl/kem"
+	"circl/kem/hybrid"
+)
+
+// Either ecdheParameters or kem.PrivateKey
+type clientKeySharePrivate interface{}
+
+var (
+	kyber512X25519CurveID = CurveID(0xFF01)
+	invalidCurveID        = CurveID(0xFFFF)
+)
+
+func kemSchemeKeyToCurveID(s kem.Scheme) CurveID {
+	switch s.Name() {
+	case "Kyber512-X25519":
+		return kyber512X25519CurveID
+	default:
+		return invalidCurveID
+	}
+}
+
+// Extract CurveID from clientKeySharePrivate
+func clientKeySharePrivateCurveID(ks clientKeySharePrivate) CurveID {
+	switch v := ks.(type) {
+	case kem.PrivateKey:
+		return kemSchemeKeyToCurveID(v.Scheme())
+	case ecdheParameters:
+		return v.CurveID()
+	default:
+		panic("cfkem: internal error: unknown clientKeySharePrivate")
+	}
+}
+
+// Returns scheme by CurveID if supported by Circl
+func curveIdToCirclScheme(id CurveID) kem.Scheme {
+	if id == kyber512X25519CurveID {
+		return hybrid.Kyber512X25519()
+	}
+	return nil
+}
+
+// Generate a new shared secret and encapsulates it for the packed
+// public key in ppk using randomness from rnd.
+func encapsulateForKem(scheme kem.Scheme, rnd io.Reader, ppk []byte) (
+	ct, ss []byte, internalErr bool, err error) {
+	pk, err := scheme.UnmarshalBinaryPublicKey(ppk)
+	if err != nil {
+		return nil, nil, false, err
+	}
+	seed := make([]byte, scheme.EncapsulationSeedSize())
+	if _, err := io.ReadFull(rnd, seed); err != nil {
+		return nil, nil, true, err
+	}
+	ct, ss, err = scheme.EncapsulateDeterministically(pk, seed)
+	return ct, ss, false, err
+}
+
+// Generate a new keypair using randomness from rnd.
+func generateKemKeyPair(scheme kem.Scheme, rnd io.Reader) (
+	kem.PublicKey, kem.PrivateKey, error) {
+	seed := make([]byte, scheme.SeedSize())
+	if _, err := io.ReadFull(rnd, seed); err != nil {
+		return nil, nil, err
+	}
+	pk, sk := scheme.DeriveKeyPair(seed)
+	return pk, sk, nil
+}
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -31,9 +31,9 @@ import (
 type clientKeySharePrivate interface{}
 
 var (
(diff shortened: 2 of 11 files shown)


src/circl/kem/hybrid/hybrid.go
```diff
@@ -80,37 +95,36 @@ func (sch *scheme) Name() string { return sch.name }
...
 func (sch *scheme) SeedSize() int {
 	first := sch.first.SeedSize()
 	second := sch.second.SeedSize()
 	ret := second
 	if first > second {
 		ret = first
 	}
-	if ret > 32 {
-		panic("SeedSize too big for SHAKE256")
-	}
```
Lekensteyn: These changes seems manually imported from https://github.com/cloudflare/circl/blob/main/kem/hybrid/hybrid.go, could you try and import the latest version by running the cf-copy-from-circle.py script?

Preferably as a separate commit so we can squash it with the `circl: import crypto library CIRCL` commit later.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -279,23 +280,38 @@ GroupSelection:
...
-		return err
+	if kem := curveIdToCirclScheme(selectedGroup); kem != nil {
+		ct, ss, internalErr, err := encapsulateForKem(kem, c.config.rand(), clientKeyShare.data)
+		if err != nil {
+			if internalErr {
+				c.sendAlert(alertInternalError)
+			} else {
+				c.sendAlert(alertIllegalParameter)
+			}
+			return err
```
Lekensteyn: Can we review all newly added error paths, and check whether the error messages (when hit) are sufficiently unique to recognize where something went wrong? And if not, maybe add more context?

For example, `encapsulateForKem` can return `internalErr == true` if reading from the randomness source fails, it could result in something like `EOF`. Presumably that is such a unique case that we can ignore.

However if the client provided some crap and `UnmarshalBinaryPublicKey` returns a failure, what kind of errors can we expect there? I think it would be something like `wrong size for public key` for `ErrPubKeySize`. If that is the case, can we use something like `fmt.Errorf("tls: kem encapsulate failed: %v", err)`

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -535,10 +559,16 @@ func (hs *clientHandshakeStateTLS13) processServerHello() error {
...
+	var sharedKey []byte
+	if params, ok := hs.keySharePrivate.(ecdheParameters); ok {
+		sharedKey = params.SharedKey(hs.serverHello.serverShare.data)
+	} else if sk, ok := hs.keySharePrivate.(circlKem.PrivateKey); ok {
+		var err error
+		sharedKey, err = sk.Scheme().Decapsulate(sk, hs.serverHello.serverShare.data)
+		if err != nil {
+			c.sendAlert(alertInternalError)
+			return err
+		}
```
Lekensteyn, line 573: Defensive programming: add an else case to `establishHandshakeKeys` in case `hs.keySharePrivate` has an unexpected type

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -347,21 +350,40 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
 			c.sendAlert(alertIllegalParameter)
 			return errors.New("tls: server selected unsupported group")
 		}
-		if hs.ecdheParams.CurveID() == curveID {
+		if clientKeySharePrivateCurveID(hs.keySharePrivate) == curveID {
 			c.sendAlert(alertIllegalParameter)
 			return errors.New("tls: server sent an unnecessary HelloRetryRequest key_share")
 		}
 		if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
```
Lekensteyn: This check probably needs to be removed as it is duplicated for the `ecdhe` case below.

Due to this check, the `if scheme := curveIdToCirclScheme(curveID); scheme != nil {` check will not be reached either, so this code never seemed to be hit. Do we have a gap in our tests that need to be addressed? I only saw some ECH test failures, not other ones:
```
--- FAIL: TestECHHandshake (0.05s)
    --- FAIL: TestECHHandshake/11 (0.00s)
        ech_test.go:746: hrr / accepted
        ech_test.go:751: client aborts; want success
        ech_test.go:755: server aborts; want success
        ech_test.go:761: client err: remote error: tls: illegal parameter
        ech_test.go:767: server err: tls: client sent invalid key share in second ClientHello
```

src/crypto/tls/cfkem.go
```diff
@@ -0,0 +1,88 @@
+// Glue to add Circl's (post-quantum) hybrid KEMs to quic-go.
```
Lekensteyn: quic-go, hmmm...

Lekensteyn (changes requested):

src/circl/kem/hybrid/hybrid.go
chris-wood: Resolved via #115.

src/crypto/tls/handshake_client.go
```diff
@@ -131,19 +131,34 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
 		}
 
 		curveID := config.curvePreferences()[0]
-		if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
-			return nil, nil, errors.New("tls: CurvePreferences includes unsupported curve")
-		}
-		params, err = generateECDHEParameters(config.rand(), curveID)
-		if err != nil {
-			return nil, nil, err
+		if scheme := curveIdToCirclScheme(curveID); scheme != nil {
```
cjpatton, line 134: Should we make sure this path is only taken if the client supports TLS 1.3?

chris-wood: Yeah, probably. I'll add a check.

src/crypto/tls/handshake_client_tls13.go
chris-wood: Good catch. This check was removed, but the bug persists, which suggests it's elsewhere.

src/crypto/tls/handshake_client_tls13.go
chris-wood: Done!

src/crypto/tls/handshake_client.go
bwesterb: This is within the `if hello.supportedVersions[0] == VersionTLS13` check.

bwesterb:
I think you need this:

```
diff --git a/src/crypto/tls/handshake_client_tls13.go b/src/crypto/tls/handshake_client_tls13.go
index fa8675f0c1..f8e35797d0 100644
--- a/src/crypto/tls/handshake_client_tls13.go
+++ b/src/crypto/tls/handshake_client_tls13.go
@@ -370,7 +370,7 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
                                return err
                        }
                        hs.keySharePrivate = sk
-                       hs.hello.keyShares = []keyShare{{group: curveID, data: packedPk}}
+                       hello.keyShares = []keyShare{{group: curveID, data: packedPk}}
                } else {
                        if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
                                c.sendAlert(alertInternalError)
@@ -382,7 +382,7 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
                                return err
                        }
                        hs.keySharePrivate = params
-                       hs.hello.keyShares = []keyShare{{group: curveID, data: params.PublicKey()}}
+                       hello.keyShares = []keyShare{{group: curveID, data: params.PublicKey()}}
                }
        }
```

src/crypto/tls/cfkem.go
cjpatton: New files should include Cloudflare copyright.

src/crypto/tls/cfkem.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"circl/kem"
+	"circl/kem/hybrid"
+)
+
+// Either ecdheParameters or kem.PrivateKey
+type clientKeySharePrivate interface{}
+
+var (
+	kyber512X25519CurveID = CurveID(0xFF01)
```
cjpatton: Is this codepoint defined in a draft, e.g., Stebila's hybrid KEX draft? If so we should point to it.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -613,6 +629,7 @@ func (hs *serverHandshakeStateTLS13) doHelloRetryRequest(selectedGroup CurveID)
 
 	if len(clientHello.keyShares) != 1 || clientHello.keyShares[0].group != selectedGroup {
 		c.sendAlert(alertIllegalParameter)
+		fmt.Println("got", clientHello.keyShares[0].group, "want", selectedGroup)
```
bwesterb: Remove debug statement

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -347,21 +350,40 @@ func (hs *clientHandshakeStateTLS13) processHelloRetryRequest() error {
...
-			return errors.New("tls: CurvePreferences includes unsupported curve")
-		}
-		params, err := generateECDHEParameters(c.config.rand(), curveID)
-		if err != nil {
-			c.sendAlert(alertInternalError)
-			return err
+		// if _, ok := curveForCurveID(curveID); curveID != X25519 && !ok {
+		// 	c.sendAlert(alertInternalError)
+		// 	return errors.New("tls: CurvePreferences includes unsupported curve")
+		// }
```
bwesterb: Remove this.

bwesterb (changes requested):

src/crypto/tls/conn.go
```diff
@@ -95,6 +95,9 @@ type Conn struct {
 	// clientProtocol is the negotiated ALPN protocol.
 	clientProtocol string
 
+	// selectedGroup is the negotiated group
```
cjpatton: This parameter is only consumed by tests in cfkem_test.go. Is there any way to test the same thing without needing to make this change? This would help reduce maintenance cost. Some potential options:
- ConnectionState.CipherSuite (not sure how we compute the cipher suite when the new code point is negotiated)
- ClientHelloInfo.SupportedCurves

src/crypto/tls/handshake_client.go
```diff
@@ -122,7 +122,7 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, ecdheParamet
 		hello.supportedSignatureAlgorithms = config.supportedSignatureAlgorithms()
 	}
 
-	var params ecdheParameters
+	var secret clientKeySharePrivate
```
cjpatton, line 125: Since we're (unavoidably) overloading the semantics of `curveID` to include KEMs, perhaps we could keep this name as well for consistency.
```suggestion
	var params clientKeySharePrivate
```

src/crypto/tls/handshake_client_tls13.go
cjpatton: nit: remove this commented out code, as it doesn't seem to serve a purpose.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -16,19 +16,22 @@ import (
...
+	ctx             context.Context
+	serverHello     *serverHelloMsg
+	hello           *clientHelloMsg
+	helloInner      *clientHelloMsg
+	keySharePrivate clientKeySharePrivate
+
+	session       *ClientSessionState
+	earlySecret   []byte
+	binderKey     []byte
+	selectedGroup CurveID
```
cjpatton, line 34: Similar comment here: If tests in`cfkem_test.go` can be implemented another way, it would be best if we can remove this new field.

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -535,10 +559,19 @@ func (hs *clientHandshakeStateTLS13) processServerHello() error {
...
-		c.sendAlert(alertIllegalParameter)
-		return errors.New("tls: invalid server key share")
+	var sharedKey []byte
+	if params, ok := hs.keySharePrivate.(ecdheParameters); ok {
+		sharedKey = params.SharedKey(hs.serverHello.serverShare.data)
+	} else if sk, ok := hs.keySharePrivate.(circlKem.PrivateKey); ok {
+		var err error
+		sharedKey, err = sk.Scheme().Decapsulate(sk, hs.serverHello.serverShare.data)
+		if err != nil {
+			c.sendAlert(alertInternalError)
```
cjpatton: Is an internal error alert appropriate? What if the failure is due to the server's input being malformed?

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -535,10 +559,19 @@ func (hs *clientHandshakeStateTLS13) processServerHello() error {
...
+			return err
+		}
+	} else {
+		c.sendAlert(alertInternalError)
+		return fmt.Errorf("tls: failed to establish handshake keys")
```
cjpatton: I think the previous error handling is more appropriate:
```suggestion
		c.sendAlert(alertIllegalParameter)
		return errors.New("tls: invalid server key share")
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -279,23 +280,38 @@ GroupSelection:
 		clientKeyShare = &hs.clientHello.keyShares[0]
 	}
 
-	if _, ok := curveForCurveID(selectedGroup); selectedGroup != X25519 && !ok {
+	if _, ok := curveForCurveID(selectedGroup); selectedGroup != X25519 && curveIdToCirclScheme(selectedGroup) == nil && !ok {
```
cjpatton, line 283: For consistency with the rest of the PR, why not do the CIRCL KEM check first, then the standard DH logic:
```go
if kem := curveIdtoCirclScheme(selectedGroup); kem != nil {
    // do CIRCL
} else if _, ok := curveForCurveID(selectedGroup); selectedGroup != X25519 && !ok {
   // do standard thing
}
```

src/crypto/tls/cfkem.go
```diff
@@ -0,0 +1,88 @@
...
+func curveIdToCirclScheme(id CurveID) kem.Scheme {
+	if id == kyber512X25519CurveID {
+		return hybrid.Kyber512X25519()
+	}
+	return nil
+}
+
+// Generate a new shared secret and encapsulates it for the packed
+// public key in ppk using randomness from rnd.
+func encapsulateForKem(scheme kem.Scheme, rnd io.Reader, ppk []byte) (
```
cjpatton, line 79: One of the return values is a flag indicating whether the error, if present, should be treated as an internal error. I find this rather unintuitive, and it led me to misread the code where this function is called. One possible alterantive would be to return both the error to propagate to the caller *and* the TLS alert to send to the peer. This pattern is used elsewhere in crypto/tls.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -279,23 +280,38 @@ GroupSelection:
...
+		params, err := generateECDHEParameters(c.config.rand(), selectedGroup)
+		if err != nil {
+			c.sendAlert(alertInternalError)
+			return err
+		}
+		hs.hello.serverShare = keyShare{group: selectedGroup, data: params.PublicKey()}
+		hs.sharedKey = params.SharedKey(clientKeyShare.data)
 	}
-	hs.hello.serverShare = keyShare{group: selectedGroup, data: params.PublicKey()}
-	hs.sharedKey = params.SharedKey(clientKeyShare.data)
```
cjpatton, line 292: nit: It would be nice if the code could be written in such a way that these two lines would not need to be removed and duplicated above, thus reducing the risk of a conflict during rebase.

src/crypto/tls/handshake_server_tls13.go
cjpatton: Remove this debug print statement

src/crypto/tls/key_agreement.go
```diff
@@ -168,7 +168,7 @@ type ecdheKeyAgreement struct {
 func (ka *ecdheKeyAgreement) generateServerKeyExchange(config *Config, cert *Certificate, clientHello *clientHelloMsg, hello *serverHelloMsg) (*serverKeyExchangeMsg, error) {
 	var curveID CurveID
 	for _, c := range clientHello.supportedCurves {
-		if config.supportsCurve(c) {
+		if config.supportsCurve(c) && curveIdToCirclScheme(c) == nil {
```
cjpatton, line 171: Might be worth testing that the new codepoint is properly ignored when TLS < 1.3 is negotiated.

cjpatton (commented):
* I suggest renaming `cfkem.go` to `cf_kem.go`. Alternatively, since there aren't that many new types/methods here, this file could simply be merged into `cf.go`. (Similarly for `cfkem_test.go`.) 
* This patch uses naming conventions that are inconsistent with the rest of the package. In particular, acronyms are supposed to be capitalized when not the first word. (E.g., "curveIDToCIRCLScheme" instead of "curveIdToCirclScheme".)
* We have happy-path tests for negotiation of the new key_shares. We should also add tests that confirm expected behavior when one party supports hybrid KEX but the other doesn't.

src/crypto/tls/cfkem.go
bwesterb: there is no codepoint yet

bwesterb (approved):

bwesterb:
@chris-wood A heads-up: I've added Kyber768 into your branch.

src/crypto/tls/handshake_server_tls13.go
bwesterb: Wrapped errors.

src/crypto/tls/cfkem.go
bwesterb: Done

src/crypto/tls/conn.go
bwesterb: We will need this for metrics later anyway.

src/crypto/tls/handshake_client.go
bwesterb: I prefer to keep the change as it's consistent with the patch that we maintain in go-boring-pq.

src/crypto/tls/handshake_client_tls13.go
bwesterb: Yeah alertIllegalParameter is better. Done.

src/crypto/tls/handshake_client_tls13.go
bwesterb: Agree. Done.

src/crypto/tls/handshake_server_tls13.go
bwesterb: I find the current diff easier to read

src/crypto/tls/cfkem.go
bwesterb: Done.

src/crypto/tls/handshake_server_tls13.go
bwesterb: I don't think there is a perfect solution here

src/crypto/tls/key_agreement.go
bwesterb: CurveID isn't a thing in TLS 1.2, is it?

bwesterb:
> * I suggest renaming `cfkem.go` to `cf_kem.go`.

That suggests that `kem` is a build tag, which it isn't.

> Alternatively, since there aren't that many new types/methods here, this file could simply be merged into `cf.go`. (Similarly for `cfkem_test.go`.)

I prefer a separate file so it's easier to maintain across different forks.

> * This patch uses naming conventions that are inconsistent with the rest of the package. In particular, acronyms are supposed to be capitalized when not the first word. (E.g., "curveIDToCIRCLScheme" instead of "curveIdToCirclScheme".)

Go itself isn't consistent either. I'll try to be a bit more consistent.

> * We have happy-path tests for negotiation of the new key_shares. We should also add tests that confirm expected behavior when one party supports hybrid KEX but the other doesn't.

Will add.

src/crypto/tls/conn.go
bwesterb: Changed it to use CFEvent API.

bwesterb:
> Will add.

Done.

src/crypto/tls/cfkem.go
Lekensteyn: Still says `quic-go` here and `qtls.X25519` below. Can we update this for our normal crypto/tls fork?

src/crypto/tls/handshake_client.go
```diff
@@ -230,16 +247,16 @@ func (c *Conn) clientHandshake(ctx context.Context) (err error) {
...
-			session:     session,
-			earlySecret: earlySecret,
-			binderKey:   binderKey,
-			hsTimings:   hsTimings,
+			c:               c,
+			ctx:             ctx,
+			serverHello:     serverHello,
+			hello:           hello,
+			helloInner:      helloInner,
+			keySharePrivate: ecdheParams,
```
Lekensteyn, line 255: Note that we still have `ecdheParams` above even though it can be either the regular `ecdheParams` type, or the PQ KEX one. Is this intentional or should it be renamed to `keySharePrivate` as well?
```
	helloBase, ecdheParams, err := c.makeClientHello(minVersion)
```

src/crypto/tls/cfkem.go
```diff
@@ -0,0 +1,98 @@
...
+	"circl/kem/hybrid"
+)
+
+// Either ecdheParameters or kem.PrivateKey
+type clientKeySharePrivate interface{}
+
+var (
+	kyber512X25519CurveID = CurveID(0xFF01)
+	kyber768X25519CurveID = CurveID(0xFF02)
+	invalidCurveID        = CurveID(0xFFFF)
```
Lekensteyn: ID 65535 (0xFFFF) is not reserved, should we use the special value 0 instead or would that have other implications?
```suggestion
	invalidCurveID        = CurveID(0x0000)
```
"The value 0 (0x0000) has been marked as reserved." from https://www.rfc-editor.org/rfc/rfc8447.html#section-9

src/crypto/tls/cfkem.go
Lekensteyn: Are 0xFF01 and 0xFFF02 arbitrary chosen? There is a conflict with existing parameters https://www.iana.org/assignments/tls-parameters/tls-parameters.xhtml#tls-parameters-8
```
65281 	arbitrary_explicit_prime_curves 	Y 	N 	[[RFC8422](https://www.iana.org/go/rfc8422)] 	
65282 	arbitrary_explicit_char2_curves 	Y 	N 	[[RFC8422](https://www.iana.org/go/rfc8422)]
```
These are `deprecated(0xFF01..0xFF02),` in https://www.rfc-editor.org/rfc/rfc8422.html#page-12

Maybe we can pick something from the "65024-65279  Reserved for Private Use" range? (`reserved (0xFE00..0xFEFF)`)

src/crypto/tls/key_agreement.go
Lekensteyn: CurveID (used in`supported_groups`, formerly `elliptic_curves`) is used in TLS 1.2. I think you are thinking about the key exchange (RSA vs ECDHE vs ...) which is part of the cipher suite in TLS 1.2.

Lekensteyn (approved):
The tests cover the desired permutations of the Curve ID being advertised at the client/server with TLS 1.2 enabled/disabled. Looks good to me!

The codepoints are now also not colliding with others, reducing the risk of failures.

PR #115, opened 2022-07-06
Fix CIRCL integration and dependencies.

armfazh:
Changes: 
 - Updated the ./cf-copy-from-circle.py script to fetch from main.
 - I added CIRCL dependencies as vendored. (Following this pattern, then it could be easier to integrate CIRCL as a vendored dep. I will investigate this option later).
 - A test script for PQ certificates was updated.


Lekensteyn:
> Following this pattern, then it could be easier to integrate CIRCL as a vendored dep. I will investigate this option later.

Is circl intended for internal use only, or are there any types to be exposed to user applications? Adding it as vendored dependency could make that harder right? I forgot the motivation for using one over another, maybe because there were extra files that we were trying to remove.

> A test script for PQ certificates was updated.

What was the reason for this? Did the old cert expire or something?

Lekensteyn (approved):
LGTM.

I saw some `golang.org/x/sys/cpu` imports in the vendored copies which could potentially be replaced by `internal/cpu`, but there are other existing instances as well, so let's keep it.

bwesterb (approved):
Thanks!

armfazh:
> > Following this pattern, then it could be easier to integrate CIRCL as a vendored dep. I will investigate this option later.
> 
> Is circl intended for internal use only, or are there any types to be exposed to user applications? Adding it as vendored dependency could make that harder right? I forgot the motivation for using one over another, maybe because there were extra files that we were trying to remove.

You raised a good question. One way to avoid that anyone imports "circl/<PACKAGE>" while using this fork is to move it as an internal stdlib package. However, CIRCL is already a public package.

> > A test script for PQ certificates was updated.
> 
> What was the reason for this? Did the old cert expire or something?

Certificate was created with dilitihum from the Round 2.

PR #116, opened 2022-07-08
Add Kyber768X25519

bwesterb:
Updates circl to branch 768x25519


cjpatton (approved):

armfazh:
does the update of the CIRCL files was done using the python script? if you have found issues with it, let me know so we can look for an easier process.

bwesterb:
> does the update of the CIRCL files was done using the python script? if you have found issues with it, let me know so we can look for an easier process.

yeah, I used the python script with updated branch variable

PR #122, opened 2022-07-21
tls: rename Kyber{512,768}X25519

bwesterb:


src/crypto/tls/cfkem.go
```diff
@@ -31,17 +31,17 @@ import (
 type clientKeySharePrivate interface{}
 
 var (
-	Kyber512X25519 = CurveID(0xfe30)
-	Kyber768X25519 = CurveID(0xfe31)
+	Kyber512X25519Draft1 = CurveID(0xfe30)
+	Kyber768X25519Draft1 = CurveID(0xfe31)
```
cjpatton: * Does `Draft1` refer to an I-D that defines these code points? If so we should add a link to the draft here.
* Typically we would write this `Draf01`.

cjpatton (changes requested):

src/crypto/tls/cfkem.go
chris-wood: +1 -- it would be great if these were actually the codepoint values advertised in [the hybrid KEX spec](https://datatracker.ietf.org/doc/draft-ietf-tls-hybrid-design/). @bwesterb, could you please reach out to Douglas et al. to see if they're willing to add these?

bwesterb: Agreed with Douglas on the name.

cjpatton (approved):

PR #126, opened 2022-08-12
tls13: send alert on invalid ECDH share

bwesterb:
This was a regression compared to upstream.

### DIFF
--- a/src/crypto/tls/handshake_client_tls13.go
+++ b/src/crypto/tls/handshake_client_tls13.go
@@ -571,7 +571,9 @@ func (hs *clientHandshakeStateTLS13) establishHandshakeKeys() error {
 			c.sendAlert(alertIllegalParameter)
 			return fmt.Errorf("%s decaps: %w", sk.Scheme().Name(), err)
 		}
-	} else {
+	}
+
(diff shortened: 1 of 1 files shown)


cjpatton (approved):

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -571,7 +571,9 @@ func (hs *clientHandshakeStateTLS13) establishHandshakeKeys() error {
 			c.sendAlert(alertIllegalParameter)
 			return fmt.Errorf("%s decaps: %w", sk.Scheme().Name(), err)
 		}
-	} else {
+	}
+
+	if sharedKey == nil {
```
Lekensteyn, line 576: This only needs to be part of the `ecdheParameters` branch, but it does not hurt to keep it here in order to reduce the delta with upstream.

Lekensteyn (approved):

src/crypto/tls/handshake_client_tls13.go
bwesterb: Indeed

Issue #130, opened 2022-08-16
DC tests don't validate certificates


jhoyla:
In `delegated_credentials_test.go` `TestDCHandshakeServerAuth` sets `clientConfig.InsecureSkipVerify = true`. Because of how golang handles global state this is propagated to all later tests. This leads to test failures if the tests are run in a different order or individually.
This means that even if you put bit flips into the test certificate signatures the tests will still pass.

claucece:
This was fixed in the kemtls branch. There was also another bug that was also fixed on that branch (I can't remember it anymore as it was a year ago). I also changed the API to something way nicer there, so that is def the code to use. It never got merged due to the debate of if kemtls should be in the main branch or not.

claucece:
I'll take a look tomorrow and compare the code.. and let you know if there is a bug that needs solving in main branch.

Issue #129, opened 2022-08-16
`getSignatureAlgorithm` returns based on the intermediate certificate


jhoyla:
`getSignatureAlgorithm` should return the signature algorithm of a certificate, but returns the signature algorithm of the certificate's parent (the intermediate cert.)
It also doesn't support RSA leaf certificates at all.

claucece:
Beautiful! Thank you for this fix!

Issue #128, opened 2022-08-16
`selectSignatureSchemeDC` does not account for the client's `supported_algorithms`


jhoyla:
`selectSignatureSchemeDC` only checks the client's `supportedSignatureAlgorithmsDC` field. Per the [spec](https://datatracker.ietf.org/doc/html/draft-ietf-tls-subcerts-15#section-4.1.1) the selected signature scheme MUST be in both the `supportedSignatureAlgorithmsDC` and `supportedSignatureAlgorithms` fields.

claucece:
YESSSS!!! Thank you for this!!

Issue #127, opened 2022-08-16
signatureSchemeForDelegatedCredential is unneccessary


jhoyla:
`signatureSchemeForDelegatedCredential` returns the list of supported signature schemes for a delegated credential; however, per the spec, DCs are explicitly restricted to a single `SignatureAlgorithm` defined in the `dc_cert_verify_algorithm` field.
We can simply use `dc_cert_verify_algorithm`, rather than call this function.

claucece:
Def agree. It was requested in a PR review, but it is better to unify them.

PR #131, opened 2022-08-16
Fix DC implementation

jhoyla:
This PR fixes a number of bugs in the DC implementation.
Specifically this addresses Issues #127, #128, #129, and #130. 
It also adds `generate_delegated_credential.go` which provides a tool matching `generate_cert.go` in style, and producing a delegated credential of the specified properties.

### DIFF
--- a/src/crypto/tls/auth.go
+++ b/src/crypto/tls/auth.go
@@ -259,72 +259,26 @@ func signatureSchemesForCertificate(version uint16, cert *Certificate) []Signatu
 	return sigAlgs
 }
 
-// signatureSchemeForDelegatedCredential returns the list of supported
-// SignatureSchemes for a given delegated credential, based on the public
-// key, and optionally filtered by its explicit
-// SupportedSignatureAlgorithmsDC.
-//
-// This function must be kept in sync with supportedSignatureAlgorithmsDC.
-func signatureSchemeForDelegatedCredential(version uint16, dc *DelegatedCredential) []SignatureScheme {
-	pub := dc.cred.publicKey
-
-	var sigAlgs []SignatureScheme
-	switch pub.(type) {
-	case *ecdsa.PublicKey:
-		pk, ok := pub.(*ecdsa.PublicKey)
-		if !ok {
-			return nil
-		}
-		switch pk.Curve {
-		case elliptic.P256():
-			sigAlgs = []SignatureScheme{ECDSAWithP256AndSHA256}
-		case elliptic.P384():
-			sigAlgs = []SignatureScheme{ECDSAWithP384AndSHA384}
-		case elliptic.P521():
-			sigAlgs = []SignatureScheme{ECDSAWithP521AndSHA512}
-		default:
-			return nil
-		}
-	case ed25519.PublicKey:
-		sigAlgs = []SignatureScheme{Ed25519}
-	case circlSign.PublicKey:
-		pk, ok := pub.(circlSign.PublicKey)
-		if !ok {
-			return nil
-		}
-		scheme := pk.Scheme()
-		tlsScheme, ok := scheme.(circlPki.TLSScheme)
-		if !ok {
-			return nil
-		}
-		sigAlgs = []SignatureScheme{SignatureScheme(tlsScheme.TLSIdentifier())}
-	default:
-		return nil
-	}
-
-	return sigAlgs
-}
-
 // selectSignatureSchemeDC picks a SignatureScheme from the peer's preference list
 // that works with the selected delegated credential. It's only called for protocol
 // versions that support delegated credential, so TLS 1.3.
-func selectSignatureSchemeDC(vers uint16, dc *DelegatedCredential, peerAlgs []SignatureScheme) (SignatureScheme, error) {
+func selectSignatureSchemeDC(vers uint16, dc *DelegatedCredential, peerAlgs []SignatureScheme, peerAlgsDC []SignatureScheme) (SignatureScheme, error) {
 	if vers != VersionTLS13 {
 		return 0, errors.New("unsupported TLS version for dc")
 	}
 
-	supportedAlgs := signatureSchemeForDelegatedCredential(vers, dc)
-	if len(supportedAlgs) == 0 {
-		return 0, errors.New("unsupported scheme for dc")
+	if !isSupportedSignatureAlgorithm(dc.algorithm, peerAlgs) {
+		return undefinedSignatureScheme, errors.New("tls: peer doesn't support the delegated credential's signature")
 	}
+
 	// Pick signature scheme in the peer's preference order, as our
 	// preference order is not configurable.
-	for _, preferredAlg := range peerAlgs {
-		if isSupportedSignatureAlgorithm(preferredAlg, supportedAlgs) {
+	for _, preferredAlg := range peerAlgsDC {
+		if preferredAlg == dc.cred.expCertVerfAlgo {
 			return preferredAlg, nil
 		}
 	}
-	return 0, errors.New("tls: peer doesn't support any of the delegated credential's signature algorithms")
+	return 0, errors.New("tls: peer doesn't support the delegated credential's signature algorithm")
 }
 
 // selectSignatureScheme picks a SignatureScheme from the peer's preference list
--- a/src/crypto/tls/delegated_credentials.go
+++ b/src/crypto/tls/delegated_credentials.go
@@ -21,6 +21,7 @@ import (
 	"crypto/ed25519"
 	"crypto/elliptic"
 	"crypto/rand"
+	"crypto/rsa"
 	"crypto/x509"
 	"encoding/binary"
 	"errors"
@@ -38,6 +39,10 @@ const (
 	dcMaxTTL          = time.Duration(dcMaxTTLSeconds * time.Second)
 	dcMaxPubLen       = (1 << 24) - 1 // Bytes
 	dcMaxSignatureLen = (1 << 16) - 1 // Bytes
+)
+
+const (
+	undefinedSignatureScheme SignatureScheme = 0x0000
 )
 
 var extensionDelegatedCredential = []int{1, 3, 6, 1, 4, 1, 44363, 44}
@@ -61,7 +66,6 @@ func isValidForDelegation(cert *x509.Certificate) bool {
 			return true
 		}
 	}
-
 	return false
 }
 
@@ -133,6 +137,7 @@ func (cred *credential) marshalPublicKeyInfo() ([]byte, error) {
 		}
 
 		return rawPub, nil
+
 	default:
 		return nil, fmt.Errorf("tls: unsupported signature scheme: 0x%04x", cred.expCertVerfAlgo)
 	}
@@ -228,6 +233,12 @@ func getHash(scheme SignatureScheme) crypto.Hash {
 		return crypto.SHA512
 	case Ed25519:
 		return directSigning
+	case PKCS1WithSHA256, PSSWithSHA256:
+		return crypto.SHA256
+	case PSSWithSHA384:
+		return crypto.SHA384
+	case PSSWithSHA512:
+		return crypto.SHA512
 	default:
 		return 0 //Unknown hash function
 	}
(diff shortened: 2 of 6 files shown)


src/crypto/tls/delegated_credentials.go
```diff
@@ -296,20 +303,22 @@ func getSignatureAlgorithm(cert *Certificate) (SignatureScheme, error) {
...
-		} else if certAlg == x509.ECDSAWithSHA512 && curveName == "P-521" {
+		} else if certAlg == x509.ECDSA && curveName == "P-521" {
 			sigAlgo = ECDSAWithP521AndSHA512
 		} else {
 			return SignatureScheme(0x00), fmt.Errorf("using curve %s for %s is not supported", curveName, cert.Leaf.SignatureAlgorithm)
 		}
 	case ed25519.PrivateKey:
 		sigAlgo = Ed25519
+	case *rsa.PrivateKey:
+		return PSSWithSHA256, nil
```
Lekensteyn, line 324: Add a comment here why this particular value is chosen, and not PSSWithSHA512 for example.

Lekensteyn (approved):
LGTM with two minor nits:
* Add a comment about why `PSSWithSHA256` is selected.
* Update the commit message with a reference to the GitHub issues being addressed.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -450,9 +450,8 @@ func cloneHash(in hash.Hash, h crypto.Hash) hash.Hash {
 // cannot be used for the current connection.
 func getDelegatedCredential(clientHello *ClientHelloInfo, cert *Certificate) (*DelegatedCredentialPair, error) {
 	if len(cert.DelegatedCredentials) == 0 {
-		return nil, errors.New("No Delegated Credential found.")
+		return nil, errors.New("no Delegated Credential found.")
```
cjpatton: nit: Why make this change? If this change is useful, consider making the corresponding change below as well.

FWIW, my linter gives the following warning:
```
error strings should not be capitalized or end with punctuation or a newline
```

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -511,16 +510,14 @@ func (hs *serverHandshakeStateTLS13) pickCertificate() error {
...
 			// Even if the Delegated Credential has already been marshalled, be sure it is the correct one.
-			delegatedCredentialPair.DC.raw, err = delegatedCredentialPair.DC.marshal()
+			delegatedCredentialPair.DC.raw, err = delegatedCredentialPair.DC.Marshal()
 			if err != nil {
 				// invalid Delegated Credential. Fallback to the certificate.
 				return nil
 			}
-
-			hs.sigAlg, err = selectSignatureSchemeDC(c.vers, delegatedCredentialPair.DC, hs.clientHello.supportedSignatureAlgorithmsDC)
+			hs.sigAlg, err = selectSignatureSchemeDC(c.vers, delegatedCredentialPair.DC, hs.clientHello.supportedSignatureAlgorithms, hs.clientHello.supportedSignatureAlgorithmsDC)
```
cjpatton: Except for setting `hs.sigAlg`, calling `selectSignatureSchemeDC()` seems to be redundant, since it's also called by `getDelegatedCredential()`. It would be more readable --- and a bit more efficient --- to plumb the signature algorithm through `getDelegatedCredential()`.

Similar comment for `getClientDelegatedCredential()`.

src/crypto/tls/auth.go
```diff
@@ -259,72 +259,26 @@ func signatureSchemesForCertificate(version uint16, cert *Certificate) []Signatu
...
+func selectSignatureSchemeDC(vers uint16, dc *DelegatedCredential, peerAlgs []SignatureScheme, peerAlgsDC []SignatureScheme) (SignatureScheme, error) {
 	if vers != VersionTLS13 {
 		return 0, errors.New("unsupported TLS version for dc")
 	}
 
-	supportedAlgs := signatureSchemeForDelegatedCredential(vers, dc)
-	if len(supportedAlgs) == 0 {
-		return 0, errors.New("unsupported scheme for dc")
+	if !isSupportedSignatureAlgorithm(dc.algorithm, peerAlgs) {
+		return SignatureScheme(0x00), errors.New("tls: peer doesn't support the delegated credential's signature")
```
cjpatton: nit: These code points are 16 bits.
```suggestion
		return SignatureScheme(0x0000), errors.New("tls: peer doesn't support the delegated credential's signature")
```

It might be a good idea to make the reserved code point an explicit variant, e.g.:

```go
const (
    undefinedSigantureScheme SignatureScheme = 0x0000
)
```

src/crypto/tls/delegated_credentials.go
```diff
@@ -296,20 +303,22 @@ func getSignatureAlgorithm(cert *Certificate) (SignatureScheme, error) {
 	case *ecdsa.PrivateKey:
 		pk := sk.Public().(*ecdsa.PublicKey)
 		curveName := pk.Curve.Params().Name
-		certAlg := cert.Leaf.SignatureAlgorithm
-		if certAlg == x509.ECDSAWithSHA256 && curveName == "P-256" {
+		certAlg := cert.Leaf.PublicKeyAlgorithm
```
cjpatton, line 309: GREAT catch. It seems like a huge oversight that the distinction between SignatureAlgorithm and PublicKeyAlgorithm is not explicitly documented. This seems easy to confuse (as we did!)

src/crypto/tls/delegated_credentials.go
```diff
@@ -296,20 +303,22 @@ func getSignatureAlgorithm(cert *Certificate) (SignatureScheme, error) {
...
 		} else {
 			return SignatureScheme(0x00), fmt.Errorf("using curve %s for %s is not supported", curveName, cert.Leaf.SignatureAlgorithm)
 		}
 	case ed25519.PrivateKey:
 		sigAlgo = Ed25519
+	case *rsa.PrivateKey:
+		return PSSWithSHA256, nil
 	default:
-		return SignatureScheme(0x00), fmt.Errorf("tls: unsupported algorithm for Delegated Credential")
+		return SignatureScheme(0x00), fmt.Errorf("tls: unsupported algorithm for signing Delegated Credential")
```
cjpatton: `undefinedSignatureScheme` would also be useful here. Alternatively, you could just return `sigAlgo`, which defaults to `0x0000`.

src/crypto/tls/delegated_credentials.go
cjpatton, line 324: nit: Following the code pattern, you would want
```suggestion
		sigAlgo = PSSWithSHA256
```

However I think what you have is actually more readable. Perhaps we should return the sigalg immediately when known everywher ein this function?

src/crypto/tls/delegated_credentials.go
```diff
@@ -448,17 +464,29 @@ func (dc *DelegatedCredential) Validate(cert *x509.Certificate, isClient bool, n
...
 		return ed25519.Verify(pk, in, dc.signature)
+	case PSSWithSHA256,
+		PSSWithSHA384,
+		PSSWithSHA512:
+		pk, ok := cert.PublicKey.(*rsa.PublicKey)
+		if !ok {
+			return false
+		}
+		hash := getHash(dc.algorithm)
+		return rsa.VerifyPSS(pk, hash, in, dc.signature, nil) == nil
```
cjpatton, line 478: Not particular to this PR, but it's a bit odd that we call the signed string `in` here but `value` above.

src/crypto/tls/delegated_credentials.go
```diff
@@ -448,17 +464,29 @@ func (dc *DelegatedCredential) Validate(cert *x509.Certificate, isClient bool, n
...
 	default:
 		return false
 	}
 }
 
 // marshal encodes a DelegatedCredential structure. It also sets dc.Raw to that
```
cjpatton:
```suggestion
// Marshal encodes a DelegatedCredential structure. It also sets dc.Raw to that
```

src/crypto/tls/generate_delegated_credential.go
```diff
@@ -0,0 +1,125 @@
...
+	signatureScheme = flag.String("signature-scheme", "", "The signature scheme used by the DC")
+	certPath        = flag.String("cert-path", "./cert.pem", "Path to signing cert")
+	keyPath         = flag.String("key-path", "./key.pem", "Path to signing key")
+	isClient        = flag.Bool("client-dc", false, "Create a client Delegated Credential")
+	outPath         = flag.String("out-path", "./", "Path to output directory")
+)
+
+var SigStringMap = map[string]tls.SignatureScheme{
+	// ECDSA algorithms. Only constrained to a specific curve in TLS 1.3.
+	"ECDSAWithP256AndSHA256": 0x0403,
```
cjpatton: Consider using https://pkg.go.dev/crypto/tls#SignatureScheme here directly, in order to reduce the chance of bugs.
```suggestion
	"ECDSAWithP256AndSHA256": tls.ECDSAWithP256AndSHA256,
```

src/crypto/tls/generate_delegated_credential.go
```diff
@@ -0,0 +1,125 @@
...
+	cert, err := tls.LoadX509KeyPair(*certPath, *keyPath)
+	if err != nil {
+		log.Fatalf("Failed to load certificate and key: %v", err)
+	}
+	cert.Leaf, err = x509.ParseCertificate(cert.Certificate[0])
+	if err != nil {
+		log.Fatalf("Failed to parse leaf certificate: %v", err)
+	}
+
+	validTime := time.Since(cert.Leaf.NotBefore) + *validFor
```
cjpatton, line 65: (At the risk of making myself look like an idiot) Is this equivalent?
```suggestion
	validTime := cert.Leaf.NotBefore.Add(*validFor)
```

src/crypto/tls/handshake_client_tls13.go
```diff
@@ -908,7 +912,7 @@ func (hs *clientHandshakeStateTLS13) sendClientCertificate() error {
 
 	if certMsg.delegatedCredential {
 		suppSigAlgo = hs.certReq.supportedSignatureAlgorithmsDC
-		sigAlgorithm, err = selectSignatureSchemeDC(c.vers, dcPair.DC, suppSigAlgo)
+		sigAlgorithm, err = selectSignatureSchemeDC(c.vers, dcPair.DC, hs.certReq.supportedSignatureAlgorithms, suppSigAlgo)
```
cjpatton: Same comment here as for `handshake_server_tls13.go`.

cjpatton (commented):
A few minor things, but otherwise this is in good shape. Thanks for your hard work on this.

src/crypto/tls/generate_delegated_credential.go
jhoyla: Sadly not. `NotBefore` is a specific/absolute time. `cert.Leaf.NotBefore.Add(*validFor)` gives a specific time.
`time.Since(cert.Leaf.NotBefore)` is the number of seconds since the cert became valid. 
`time.Since(cert.Leaf.NotBefore) + *validFor` is a duration, as required by the spec.

src/crypto/tls/delegated_credentials.go
jhoyla: I think the idea is that here `in` is an untrusted input we want to validate, but above it's a value we're preparing to sign / commit to. Not sure I particularly buy into the logic, but I think it does highlight that the function is being used in two different ways.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -511,22 +506,14 @@ func (hs *serverHandshakeStateTLS13) pickCertificate() error {
...
 			}
-
-			hs.sigAlg, err = selectSignatureSchemeDC(c.vers, delegatedCredentialPair.DC, hs.clientHello.supportedSignatureAlgorithmsDC)
-			if err != nil {
-				// the Delegated Credential is unsupported or
-				// incompatible with the client's signature algorithms.
-				// Fallback to the certificate.
-				return nil
-			}
+			hs.sigAlg = delegatedCredentialPair.DC.cred.expCertVerfAlgo
```
cjpatton, line 517: Nice. Note: `selectSignatureSchemeDC` checks that that support for `dc_expected_cert_verify_algorithm` was indicated by the client. I think it would be worth documenting this fact above `getDelegatedCredential()` so that it's clear to the reader.

cjpatton (approved):

src/crypto/tls/common.go
```diff
@@ -442,6 +442,9 @@ type ClientSessionCache interface {
 // RFC 8446, Section 4.2.3.
 type SignatureScheme uint16
 
+const (
+	undefinedSignatureScheme SignatureScheme = 0x0000
+)
```
cjpatton: nit: This is a potential rebase conflict risk. Consider moving to `delegated_credentials.go`.

src/crypto/tls/auth.go
```diff
@@ -340,6 +294,7 @@ func selectSignatureScheme(vers uint16, c *Certificate, peerAlgs []SignatureSche
 		// can assume that it supports SHA1. See RFC 5246, Section 7.4.1.4.1.
 		peerAlgs = []SignatureScheme{PKCS1WithSHA1, ECDSAWithSHA1}
 	}
+
```
cjpatton: nit: We should minimize changes to upstream files where possible, in order reduce rebase conflicts.

PR #132, opened 2022-09-02
tls: report CurveIDs for TLS <1.2 kexes

bwesterb:


src/crypto/tls/handshake_client.go
```diff
@@ -586,6 +586,12 @@ func (hs *clientHandshakeState) doFullHandshake() error {
 		if err != nil {
 			return err
 		}
+
```
Lekensteyn: Should we move this before the above `c.readHandshake()`?

Lekensteyn (approved):
LGTM, one minor suggestion to move some code to be closer to the relevant dependency.

src/crypto/tls/handshake_client.go
bwesterb: Done.

PR #144, opened 2023-03-01
Normal vendor, restructure and add P256Kyber768Draft00

bwesterb:
Several changes to align this fork closer to some of our other Go forks.

1. Vendor CIRCL in the normal way, instead of copying under a `circl` package and update to `main`.
2. Merge three PQ signatures commits into one.
3. Adds P256Kyber768Draft00 key agreement.
5. Left-over nits of @cjpatton  from 1.20.1 rebase

To generate range diff, use `git range-diff go1.20.1..cf/cf go1.20.1..cf/bas/restructure`. It's too long to be pasted.

PR #145, opened 2023-03-02
Fix bugs in #144

bwesterb:
`git range-diff go1.20.1..cf/cf go1.20.1..cf/bas/fix-restructure`


src/go/build/deps_test.go
```diff
@@ -656,6 +666,11 @@ func TestDependencies(t *testing.T) {
 		}
 		var bad []string
 		for _, imp := range imports {
+			// TODO Remove this exception for github.com/cloudflare/circl
```
armfazh: is this TODO comment still valid?

Lekensteyn (approved):
This fixes tests failures by bumping x/sys from v0.3.0 to v0.5.0 in src/cmd/go.sum, adds missing deps info to src/go/build/deps_test.go, updates a commit message, and also updates the vendored copy of x/sys in src/cmd/vendor/.

LGTM. Hopefully in the future we can avoid making these vendor and go.mod changes by delaying non-urgent updates of dependencies in circl.

src/go/build/deps_test.go
bwesterb: Yes.

PR #148, opened 2023-05-25
crypto/tls: add new X25519Kyber768Draft00 code point

Lekensteyn:
* Point tls.X25519Kyber768Draft00 to the new 0x6399 identifier while the old 0xfe31 identifier is available as tls.X25519Kyber768Draft00Old.
* Make sure that the kem.PrivateKey can always be mapped to the CurveID that was linked to it. This is needed since we now have two ID aliasing to the same scheme, and clients need to be able to detect whether the key share presented by the server actually matches the key share that the client originally sent.
* Update tests, add the new identifier and remove unnecessary code.

Link: https://mailarchive.ietf.org/arch/msg/tls/HAWpNpgptl--UZNSYuvsjB-Pc2k/
Link: https://datatracker.ietf.org/doc/draft-tls-westerbaan-xyber768d00/02/
___
The current solution is the cleanest I could come up with. Alternatives considered:
* Create a new implementation of the `kem.Scheme` interface, wrapping the `hybrid.Kyber768X25519()` scheme. I dropped this idea due to the amount of boiler plate code. In the sample below, I already wrapped the scheme, but I would also have to wrap the returned `kem.PrivateKey` and `kem.PublicKey` values since `clientKeySharePrivateCurveID` used their `Scheme()` method to get back the scheme name.
* Alias to the same `hybrid.Kyber768X25519()` scheme. Fails for the same reason as above: I cannot get back to the original code point that was in use.

<details>
<summary>Incomplete wrapper code sample for kem.Scheme</summary>

```go
package tls

import (
	"github.com/cloudflare/circl/kem"
	"github.com/cloudflare/circl/kem/hybrid"
)

// kemProxyScheme implements kem.Scheme, but aliasing it under a different name.
type kemProxyScheme struct {
	name   string
	scheme kem.Scheme
}

// Assert that kemProxyScheme implements kem.Scheme
var _ kem.Scheme = (*kemProxyScheme)(nil)

func (s *kemProxyScheme) Name() string { return s.name }

func (s *kemProxyScheme) GenerateKeyPair() (kem.PublicKey, kem.PrivateKey, error) {
	return s.scheme.GenerateKeyPair()
}
func (s *kemProxyScheme) Encapsulate(pk kem.PublicKey) (ct, ss []byte, err error) {
	return s.scheme.Encapsulate(pk)
}
func (s *kemProxyScheme) Decapsulate(sk kem.PrivateKey, ct []byte) ([]byte, error) {
	return s.scheme.Decapsulate(sk, ct)
}
func (s *kemProxyScheme) UnmarshalBinaryPublicKey(data []byte) (kem.PublicKey, error) {
	return s.scheme.UnmarshalBinaryPublicKey(data)
}
func (s *kemProxyScheme) UnmarshalBinaryPrivateKey(data []byte) (kem.PrivateKey, error) {
	return s.scheme.UnmarshalBinaryPrivateKey(data)
}
func (s *kemProxyScheme) CiphertextSize() int { return s.scheme.CiphertextSize() }
func (s *kemProxyScheme) SharedKeySize() int  { return s.scheme.SharedKeySize() }
func (s *kemProxyScheme) PrivateKeySize() int { return s.scheme.PrivateKeySize() }
func (s *kemProxyScheme) PublicKeySize() int  { return s.scheme.PublicKeySize() }
func (s *kemProxyScheme) DeriveKeyPair(seed []byte) (kem.PublicKey, kem.PrivateKey) {
	return s.scheme.DeriveKeyPair(seed)
}
func (s *kemProxyScheme) SeedSize() int { return s.scheme.SeedSize() }
func (s *kemProxyScheme) EncapsulateDeterministically(pk kem.PublicKey, seed []byte) (ct, ss []byte, err error) {
	return s.scheme.EncapsulateDeterministically(pk, seed)
}
func (s *kemProxyScheme) EncapsulationSeedSize() int { return s.scheme.EncapsulationSeedSize() }

var hybridKyber768X25519Old = &kemProxyScheme{
	"Kyber768-X25519Old",
	hybrid.Kyber768X25519(),
}
```
</details>

### DIFF
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -6,13 +6,12 @@
 // To enable set CurvePreferences with the desired scheme as the first element:
 //
 //   import (
-//      "github.com/cloudflare/circl/kem/tls"
-//      "github.com/cloudflare/circl/kem/hybrid"
+//      "crypto/tls"
 //
 //          [...]
 //
 //   config.CurvePreferences = []tls.CurveID{
-//      hybrid.X25519Kyber512Draft00().(tls.TLSScheme).TLSCurveID(),
+//      tls.X25519Kyber768Draft00,
 //      tls.X25519,
 //      tls.P256,
 //   }
@@ -29,38 +28,27 @@ import (
 	"github.com/cloudflare/circl/kem/hybrid"
 )
 
-// Either ecdheParameters or kem.PrivateKey
+// Either *ecdh.PrivateKey or *kemPrivateKey
 type clientKeySharePrivate interface{}
 
-var (
-	X25519Kyber512Draft00 = CurveID(0xfe30)
-	X25519Kyber768Draft00 = CurveID(0xfe31)
-	P256Kyber768Draft00   = CurveID(0xfe32)
-	invalidCurveID        = CurveID(0)
-)
-
-func kemSchemeKeyToCurveID(s kem.Scheme) CurveID {
-	switch s.Name() {
-	case "Kyber512-X25519":
-		return X25519Kyber512Draft00
-	case "Kyber768-X25519":
-		return X25519Kyber768Draft00
-	case "P256Kyber768Draft00":
-		return P256Kyber768Draft00
-	default:
-		return invalidCurveID
-	}
+type kemPrivateKey struct {
+	secretKey kem.PrivateKey
+	curveID   CurveID
 }
+
+var (
+	X25519Kyber512Draft00    = CurveID(0xfe30)
+	X25519Kyber768Draft00    = CurveID(0x6399)
+	X25519Kyber768Draft00Old = CurveID(0xfe31)
+	P256Kyber768Draft00      = CurveID(0xfe32)
+	invalidCurveID           = CurveID(0)
+)
 
 // Extract CurveID from clientKeySharePrivate
 func clientKeySharePrivateCurveID(ks clientKeySharePrivate) CurveID {
 	switch v := ks.(type) {
-	case kem.PrivateKey:
-		ret := kemSchemeKeyToCurveID(v.Scheme())
-		if ret == invalidCurveID {
-			panic("cfkem: internal error: don't know CurveID for this KEM")
-		}
-		return ret
+	case *kemPrivateKey:
+		return v.curveID
 	case *ecdh.PrivateKey:
 		ret, ok := curveIDForCurve(v.Curve())
 		if !ok {
@@ -77,7 +65,7 @@ func curveIdToCirclScheme(id CurveID) kem.Scheme {
 	switch id {
 	case X25519Kyber512Draft00:
 		return hybrid.Kyber512X25519()
-	case X25519Kyber768Draft00:
+	case X25519Kyber768Draft00, X25519Kyber768Draft00Old:
 		return hybrid.Kyber768X25519()
 	case P256Kyber768Draft00:
 		return hybrid.P256Kyber768Draft00()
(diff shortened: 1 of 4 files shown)


src/crypto/tls/cfkem_test.go
```diff
@@ -7,28 +7,16 @@ import (
...
+func testHybridKEX(t *testing.T, curveID CurveID, clientPQ, serverPQ,
 	clientTLS12, serverTLS12 bool) {
 	var clientSelectedKEX *CurveID
 	var retry bool
 
-	rsaCert := Certificate{
-		Certificate: [][]byte{testRSACertificate},
-		PrivateKey:  testRSAPrivateKey,
-	}
-	serverCerts := []Certificate{rsaCert}
```
Lekensteyn, line 24: Reviewer note: this code was redundant since `testConfig` already contains working certificates.

src/crypto/tls/cfkem_test.go
```diff
@@ -44,15 +32,13 @@ func testHybridKEX(t *testing.T, scheme kem.Scheme, clientPQ, serverPQ,
 
 	serverConfig := testConfig.Clone()
 	if serverPQ {
-		serverConfig.CurvePreferences = []CurveID{
-			kemSchemeKeyToCurveID(scheme),
-			X25519,
-		}
+		serverConfig.CurvePreferences = []CurveID{curveID, X25519}
+	} else {
+		serverConfig.CurvePreferences = []CurveID{X25519}
```
Lekensteyn, line 37: Reviewer note: if PQ is disabled, we explicitly set non-PQ curves here to handle cases where PQ support is enabled by default for the server.

src/crypto/tls/cfkem_test.go
```diff
@@ -78,44 +64,43 @@ func testHybridKEX(t *testing.T, scheme kem.Scheme, clientPQ, serverPQ,
...
 
+	if expectedRetry != retry {
+		t.Errorf("Expected retry=%v, got retry=%v", expectedRetry, retry)
+	}
 	if clientSelectedKEX == nil {
 		t.Error("No KEX happened?")
-	}
-
-	if *clientSelectedKEX != expectedKEX {
+	} else if *clientSelectedKEX != expectedKEX {
```
Lekensteyn, line 80: Reviewer note: this code was moved here to avoid a nil dereference panic when the handshake fails.

src/crypto/tls/cfkem.go
```diff
@@ -29,38 +28,27 @@ import (
...
+}
+
 var (
-	X25519Kyber512Draft00 = CurveID(0xfe30)
-	X25519Kyber768Draft00 = CurveID(0xfe31)
-	P256Kyber768Draft00   = CurveID(0xfe32)
-	invalidCurveID        = CurveID(0)
+	X25519Kyber512Draft00    = CurveID(0xfe30)
+	X25519Kyber768Draft00    = CurveID(0x6399)
+	X25519Kyber768Draft00Old = CurveID(0xfe31)
```
cjpatton, line 42:
```suggestion
	LegacyX25519Kyber768Draft00 = CurveID(0xfe31)
```

src/crypto/tls/cfkem.go
```diff
@@ -29,38 +28,27 @@ import (
...
+	secretKey kem.PrivateKey
+	curveID   CurveID
+}
+
 var (
-	X25519Kyber512Draft00 = CurveID(0xfe30)
-	X25519Kyber768Draft00 = CurveID(0xfe31)
-	P256Kyber768Draft00   = CurveID(0xfe32)
-	invalidCurveID        = CurveID(0)
+	X25519Kyber512Draft00    = CurveID(0xfe30)
```
cjpatton, line 40: Did you consider dropping Kyber512? My understanding is that (1) it's not currently in use anywhere and (2) we're not going to need it in the future.

cjpatton (approved):
Given the requirements this solution seems to me the best choice. (At least I don't have any better ideas!) Also the implementation looks good to me. That said, why do we need to maintain the old codepoint? Are there any clients that use it, perhaps Firefox or some internal service?

bwesterb (approved):

bwesterb (approved):

bwesterb (approved):

src/crypto/tls/cfkem.go
bwesterb: We deployed Kyber512 alongside Kyber768 to get data on whether it's more deployable or not.

src/crypto/tls/cfkem.go
Lekensteyn: I'll keep the `*Old` naming to remain consistent with the boringssl naming.

Lekensteyn:
> why do we need to maintain the old codepoint? Are there any clients that use it, perhaps Firefox or some internal service?

We've been advertising it in various places (blogs, etc.) and would like to keep the code samples working for some transitional period until the new kex is widely deployed.

PR #149, opened 2023-07-12
Request callback plumbing

chris-wood:
This is just a draft. Note that this only works for the HTTP/1.1 path. HTTP/2, which uses a different connection under the hood, needs some work. The gist for testing it [is here](https://gist.github.com/chris-wood/411e650f649aca72bc744ee35b55a0d8).


src/net/http/client.go
```diff
@@ -59,7 +59,7 @@ type Client struct {
 	// Transport specifies the mechanism by which individual
 	// HTTP requests are made.
 	// If nil, DefaultTransport is used.
-	Transport RoundTripper
+	Transport ExtendedRoundTripper
```
bwesterb, line 62: This is a public field. This would break any caller who doesn't pass in a RoundTripper without `RoundTripWithCallback`.

bwesterb: What is the problem you want to solve?

chris-wood: That's a fair point. We can just extend the struct rather than modify this field. That's easy enough.

PR #151, opened 2023-08-07
Implement TLSFlags extension

jhoyla:
This PR implements the first half of the [TLS Flags extension](https://datatracker.ietf.org/doc/draft-ietf-tls-tlsflags/).


bwesterb (approved):

bwesterb:
It needs some tests.

src/crypto/tls/common.go
```diff
@@ -851,6 +858,8 @@ type Config struct {
 	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
 	SupportDelegatedCredential bool
```
cjpatton, line 923: Please document here what `TLSFlagSet` is.

src/crypto/tls/common.go
```diff
@@ -462,6 +463,12 @@ const (
 	ECDSAWithSHA1 SignatureScheme = 0x0203
 )
 
+type TLSFlag uint16
+
+const (
+	FlagSupportMTLS TLSFlag = 0x50
```
cjpatton, line 485: Please document what this is. IIRC this is not yet a codepoint that's documented? If so, we should mark it as experimental in the API.
```suggestion
        /// ExperimentalFlagSuppportMTLS ...
	ExperimentalFlagSupportMTLS TLSFlag = 0x50
```

cjpatton (changes requested):
+1 to tests. We should at least test that, when the flags are enabled by the client in crypto/tls.Config, the server sees the value in the ClientHello.

Another high-level question: How does this interact with ECH?
1. Is the the extension present in both the inner and outer handshake?
2. If so, should it appear only in the inner handshake?

bwesterb:
> Another high-level question: How does this interact with ECH?

Very good question.

jhoyla:
Added code for server side.
Tests to come tomorrow.

cjpatton:
> Another high-level question: How does this interact with ECH?
> 
>     1. Is the the extension present in both the inner and outer handshake?
> 
>     2. If so, should it appear only in the inner handshake?

Currently the extension will be added to the inner handshake only: https://github.com/cloudflare/go/blob/cf/src/crypto/tls/ech.go#L81-L85

Only specific extensions are copied into the outer handshake: https://github.com/cloudflare/go/blob/cf/src/crypto/tls/ech.go#L92-L106

In my opinion, this is the correct behavior, assuming conservatively that the value of the TLS flags extension is privacy sensitive: https://www.ietf.org/archive/id/draft-ietf-tls-esni-16.html#name-outer-clienthello

Note that since it appears in the inner handshake, it will be used by the server to terminate the connection. OTOH, if ECH is rejected, then it won't be used by the server to terminate the connection.

Lekensteyn:
(Rebased on the `cf` branch based on Go 1.21.1, no other changes.)

src/crypto/tls/handshake_messages.go
```diff
@@ -577,6 +586,18 @@ func (m *clientHelloMsg) unmarshal(data []byte) bool {
...
+			if !extData.ReadUint8LengthPrefixed(&flagsList) || flagsList.Empty() {
+				return false
+			}
+			for !flagsList.Empty() {
+				var flagByte uint8
+				if !flagsList.ReadUint8(&flagByte) {
+					return false
+				}
+				m.tlsFlags = append(m.tlsFlags, flagByte)
+			}
```
Lekensteyn, line 600: I think that this whole block can be simplified to:
```go
var flagsList cryptobyte.String
if !extData.ReadUint8LengthPrefixed(&flagsList) || flagsList.Empty() {
    return false
}
m.tlsFlags = flagsList
```
That would save some unnecessary memory allocations as well by avoiding the `append`. A `cryptobyte.String` is an alias for `[]byte`: https://pkg.go.dev/golang.org/x/crypto/cryptobyte#String

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -318,6 +319,23 @@ GroupSelection:
 		c.sendAlert(alertIllegalParameter)
 		return errors.New("tls: invalid client key share")
 	}
+	if len(hs.clientHello.tlsFlags) != 0 {
```
Lekensteyn, line 324: Missing check: `hs.clientHello.tlsFlags[0] != 0` or else you need to `c.sendAlert(alertIllegalParameter)`.

To comply with: "An implementation that receives an all-zero value for this extension or a value that contains trailing zero bytes MUST generate a fatal illegal_parameter alert."

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -318,6 +319,23 @@ GroupSelection:
...
+		supportedFlags, err := encodeFlags(hs.c.config.TLSFlagsSupported)
+		if err != nil {
+			return errors.New("tls: invalid server flags")
```
Lekensteyn, line 327: Perhaps you should also send an internal error alert here?

src/crypto/tls/common.go
```diff
@@ -465,6 +466,12 @@ const (
 	ECDSAWithSHA1 SignatureScheme = 0x0203
 )
 
+type TLSFlag uint16
```
Lekensteyn, line 482: Let's make clear that this is the flag number (index) rather than the raw representation:
```suggestion
// TLSFlag represents the flag number.
type TLSFlag uint16
```

src/crypto/tls/common.go
```diff
@@ -125,6 +125,7 @@ const (
 	extensionRenegotiationInfo       uint16 = 0xff01
 	extensionECH                     uint16 = 0xfe0d // draft-ietf-tls-esni-13
 	extensionECHOuterExtensions      uint16 = 0xfd00 // draft-ietf-tls-esni-13
+	extensionTLSFlags                uint16 = 0xfe01 // draft-ietf-tls-tlsflags-12
```
Lekensteyn, line 129: 0xfe01 (65025) belongs to the "Unassigned" namespace. Unless we have it assigned, we should probably stick to the "Reserved for Private Use" namespace which has a range starting at 0xff02 (65282).

https://www.iana.org/assignments/tls-extensiontype-values/tls-extensiontype-values.xhtml#tls-extensiontype-values-1

src/crypto/tls/common.go
Lekensteyn: `TLSFlagsSupported` still needs to be documented. It appears that it controls both the client and server-supported flags.

You should document that this is currently limited to controlling TLS Flags in the Server Hello. The spec also allows EncryptedExtensions, Certificate, Hello Retry Request.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -350,6 +368,23 @@ GroupSelection:
...
 
+func decodeFlags(flagBytes []byte) ([]TLSFlag, error) {
+	var flags []TLSFlag
+	for byteIndex, b := range flagBytes {
+		for i := 0; !(b == 0); i++ {
+			if (b & 1) == 1 {
+				flagNo := byteIndex*8 + i
+				if flagNo >= int(maxTLSFlag) {
+					return nil, fmt.Errorf("TLS flag is out of range: %d", flagNo)
+				}
```
Lekensteyn, line 394: Optimization: This check can be removed by checking outside the outer loop. Fail on `len(flagBytes) > 255` or alternatively `len(flagBytes) * 8 > maxTLSFlag`.

src/crypto/tls/common.go
Lekensteyn: Is it intentional that this flag occupies 11 bytes? 0x50 is 80 decimal, and since we start counting at 0, this occupies `ceil(81 / 8) = 11` bytes.

jhoyla: Earlier versions of the TLS Flags draft offered guidance on how to select which bit you were using. Following it gave me bit 80. That guidance is now removed so I can select another bit if it's preferred.

PR #155, opened 2023-09-28
Add tls.Config.CurveGuess to allow specifying which keyshares to send

bwesterb:


src/crypto/tls/handshake_client.go
```diff
@@ -146,30 +146,38 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, clientKeySha
...
-			if err != nil {
-				return nil, nil, err
+		hello.keyShares = []keyShare{}
+		curveIDs := []CurveID{config.curvePreferences()[0]}
+
+		if config.CurveGuess != nil {
+			curveIDs = config.CurveGuess
+		}
+
+		for _, curveID := range curveIDs {
```
lukevalenta, line 157: Should we also check here that the CurveGuess list is a subset of CurvePreferences?

bwesterb: Good question. Let me check the spec.

Also we should check for duplicates.

bwesterb: Updated.

src/crypto/tls/common.go
```diff
@@ -837,6 +837,18 @@ type Config struct {
 	// which is currently TLS 1.3.
 	MaxVersion uint16
 
+	// CurveGuess contains the "curves" for which the client will create
+	// a keyshare in the initial ClientHello for TLS 1.3. If the client
+	// guesses incorrecly, and the server does not support or does not
```
SulemanAhmadd: Nit: 
```suggestion
	// guesses incorrectly, and the server does not support or does not
```

src/crypto/tls/handshake_client.go
```diff
@@ -146,30 +146,60 @@ func (c *Conn) makeClientHello(minVersion uint16) (*clientHelloMsg, clientKeySha
 			hello.cipherSuites = append(hello.cipherSuites, defaultCipherSuitesTLS13NoAES...)
 		}
 
-		curveID := config.curvePreferences()[0]
-		if scheme := curveIdToCirclScheme(curveID); scheme != nil {
-			pk, sk, err := generateKemKeyPair(scheme, curveID, config.rand())
-			if err != nil {
-				return nil, nil, fmt.Errorf("generateKemKeyPair %s: %w",
-					scheme.Name(), err)
+		hello.keyShares = []keyShare{}
```
SulemanAhmadd: We know the length of keyShares (after moving it to line 155).
```suggestion
		hello.keyShares = make([]keyShare, 0, len(curveIDs))
```

lukevalenta (approved):

SulemanAhmadd (approved):

Lekensteyn (approved):
LGTM.

Interesting detail that is not documented here: some implementations (OpenSSL) use the order of `ClientCurveGuess` (`key_shares`) to select the final key share. Others (boringssl, Go crypto/tls) determine the expected key share based on the CurvePreferences (`supported_groups`) setting.

PR #156, opened 2023-10-02
Fix cfgo build tag and add a test

bwesterb:


Lekensteyn (approved):
During the next rebase, we should replace the `cmd/go/internal/imports, go/build: add cfgo build tag` commit with it.

PR #157, opened 2023-10-02
Use server's preferred key agreement

bwesterb:
In contrast to upstream Go, we will send an HelloRetryRequest and accept an extra roundtrip if there is a more preferred group, than the one the client has provided a keyshare for in the initial ClientHello.

Cf. https://datatracker.ietf.org/doc/draft-davidben-tls-key-share-prediction/

### DIFF
--- a/src/crypto/tls/handshake_server_tls13.go
+++ b/src/crypto/tls/handshake_server_tls13.go
@@ -256,32 +256,39 @@ func (hs *serverHandshakeStateTLS13) processClientHello() error {
 		}
 	}
 
-	// Pick the ECDHE group in server preference order, but give priority to
-	// groups with a key share, to avoid a HelloRetryRequest round-trip.
+	// Pick group by server preference. In contrast to upstream Go, we will
+	// send an HelloRetryRequest and accept an extra roundtrip if there is
+	// a more preferred group, than those for which the client has sent
+	// a keyshare in the initial ClientHello.
+	// Cf. https://datatracker.ietf.org/doc/draft-davidben-tls-key-share-prediction/
 	var selectedGroup CurveID
 	var clientKeyShare *keyShare
 GroupSelection:
 	for _, preferredGroup := range supportedCurves {
-		for _, ks := range hs.clientHello.keyShares {
-			if ks.group == preferredGroup {
-				selectedGroup = ks.group
-				clientKeyShare = &ks
-				break GroupSelection
-			}
-		}
-		if selectedGroup != 0 {
-			continue
-		}
 		for _, group := range hs.clientHello.supportedCurves {
 			if group == preferredGroup {
 				selectedGroup = group
-				break
+				break GroupSelection
+			}
+		}
+
+		// supported_groups is not necessarily sent
+		for _, ks := range hs.clientHello.keyShares {
+			if ks.group == preferredGroup {
+				selectedGroup = ks.group
+				break GroupSelection
 			}
 		}
 	}
 	if selectedGroup == 0 {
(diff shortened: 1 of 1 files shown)


src/crypto/tls/handshake_server_tls13.go
```diff
@@ -256,33 +256,40 @@ func (hs *serverHandshakeStateTLS13) processClientHello() error {
...
 GroupSelection:
 	for _, preferredGroup := range supportedCurves {
+		for _, group := range hs.clientHello.supportedCurves {
+			if group == preferredGroup {
+				selectedGroup = group
+				break GroupSelection
+			}
+		}
+
+		// supported_groups is not necessarily sent
```
Lekensteyn, line 275: This may be true for TLS 1.2, but I think that we can assume that `supported_groups` is always available in TLS 1.3. The specs says:

```
  All implementations MUST send and use these extensions when offering
   applicable features:
[..]
   -  "supported_groups" is REQUIRED for ClientHello messages using DHE
      or ECDHE key exchange.
```
from https://datatracker.ietf.org/doc/html/rfc8446#section-9.2, and from https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.8

```
   [..]  Each KeyShareEntry value MUST correspond to a
   group offered in the "supported_groups" extension and MUST appear in
   the same order.
```
Reading this, I think we might have to update the previous `ClientCurveGuess` patch to enforce this ordering. If I recall correctly, the ClientCurveGuess order would always be used for key shares, ignoring CurvePreferences. (The CurvePreferences comment could also be updated while at it, it currently suggests that clients use the first preference as key share.)

If we can assume availability of the `supported_groups` extension, we can simplify this further by removing the next loop.

bwesterb: I think it's more robust not to assume the client will always send it. For one, Go's own tests don't pass because it doesn't always construct the ClientHello properly.

bwesterb: [Example](https://github.com/cloudflare/go/blob/cf/src/crypto/tls/handshake_server_test.go#L1928-L1933)

Lekensteyn: In the current Go code, if the client does not send a `supported_groups` extension at all, but receives `key_shares` for unsupported curves, the handshake will fail. If we are going to support this case of a missing `supported_groups` extension, we should fix this and default to at least X25519 and P-256 which are [mandatory to implement](https://datatracker.ietf.org/doc/html/rfc8446#section-9.1). I would however favor simpler code: do not implement code that supports spec-violating behavior, unless there are actual real, significant clients broken by this.

FWIW, boringssl also fails when no `supported_groups` is present for TLS 1.3. See `tls1_get_shared_group`.

bwesterb: Did you check the example?

bwesterb: I think it's a trade-off between the concern you have, and keeping the patch simple.

bwesterb: [This is the alternative.](https://github.com/cloudflare/go/tree/bas/server-kex-pref-2) Let me know which one you prefer.

Lekensteyn: I think I prefer the alternative for the following reasons:
- Less code.
- Removal of code that would support non-compliant implementations.
- The context here has changed so much, so it shouldn't make much of a difference in terms of merge conflict handling.

PR #159, opened 2023-10-02
Ensure that ClientCurveGuess is a subsequence of CurvePreferences

bwesterb:


src/crypto/tls/cfgo_test.go
```diff
@@ -5,5 +5,5 @@ package tls
 import "testing"
 
 func TestCfgoBuildTag(t *testing.T) {
-    t.Error("Build tag cfgo is expected to be set for this toolchain")
+	t.Error("Build tag cfgo is expected to be set for this toolchain")
```
bwesterb, line 8: ... will fix when rebasing Thursday.

Lekensteyn (approved):
Tests would be nice to have, but the code looks correct.

Interestingly while `key_shares` must be unique per group, the spec does not require the [`supported_groups`](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7) list entries to be unique. There is no code in the upstream crypto/tls library to enforce this, nor in our fork. Your change does not affect this situation.

bwesterb:
> Interestingly while `key_shares` must be unique per group, the spec does not require the [`supported_groups`](https://datatracker.ietf.org/doc/html/rfc8446#section-4.2.7) list entries to be unique. There is no code in the upstream crypto/tls library to enforce this, nor in our fork. Your change does not affect this situation.

Interesting observation.

PR #160, opened 2023-10-03
Use server's preferred key agreeement (take 2)

bwesterb:
In contrast to upstream Go, we will send an HelloRetryRequest and accept an extra roundtrip if there is a more preferred group, than the one the client has provided a keyshare for in the initial ClientHello.

Cf. https://datatracker.ietf.org/doc/draft-davidben-tls-key-share-prediction/

Supersedes #157


Lekensteyn (approved):
LGTM (maybe squash).

One caveat though: there is no way to specify equal preference. So if a client supports `{P256, X25519}` and sends key share `P256`, then a server with CurvePreferences `{X25519, P256}` will still end up with a HRR to `P256`. Is that intended?

bwesterb:
Yes, that aligns with BoringSSL's behaviour. To specify equal preference, we need a different interface.

PR #162, opened 2023-10-10
Add dummy key agreement with codepoint 0xfe33

bwesterb:
DummyKex is a key agreeement similar in size but purposefully incompatible with X25519. The goal is to have a key agreement that servers will not support, so we can test HelloRetryRquest.

### DIFF
--- a/src/crypto/tls/cfkem.go
+++ b/src/crypto/tls/cfkem.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/cloudflare/circl/hpke"
 	"github.com/cloudflare/circl/kem"
 	"github.com/cloudflare/circl/kem/hybrid"
 )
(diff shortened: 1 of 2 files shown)


src/crypto/tls/cfkem.go
```diff
@@ -58,6 +64,8 @@ func curveIdToCirclScheme(id CurveID) kem.Scheme {
 		return hybrid.Kyber768X25519()
 	case P256Kyber768Draft00:
 		return hybrid.P256Kyber768Draft00()
+	case DummyKex:
+		return hpke.KEM_X25519_HKDF_SHA256.Scheme()
```
Lekensteyn, line 68: What is the intended semantics of this? Are we expected to negotiate it? Can we have a test in cfkem_test.go for this identifier?

bwesterb: Added test. Semantics is as explained in the comment of the CurveID.

Lekensteyn: Thanks, the test at least assures that we won't crash.

Lekensteyn (approved):

PR #163, opened 2023-10-10
Send empty keyshare extension instead of leaving it out

bwesterb:


Lekensteyn (approved):
The `AddBytes()` function works with a nil parameter. LGTM.

Can you squash this in the next rebase?

lukevalenta (approved):

PR #191, opened 2025-05-06
Implement TLS Flags and the request mTLS flag.

jhoyla:
Add support for sending and receiving TLS Flags, and support for the request mTLS flag.

### DIFF
--- a/src/crypto/tls/common.go
+++ b/src/crypto/tls/common.go
@@ -126,6 +126,7 @@ const (
 	extensionRenegotiationInfo       uint16 = 0xff01
 	extensionECH                     uint16 = 0xfe0d // draft-ietf-tls-esni-13
 	extensionECHOuterExtensions      uint16 = 0xfd00 // draft-ietf-tls-esni-13
+	extensionTLSFlags                uint16 = 0xfe01 // draft-ietf-tls-tlsflags-12
 )
 
 // TLS signaling cipher suite values
@@ -364,6 +365,17 @@ type ConnectionState struct {
 	// This means the client has offered ECH or sent GREASE ECH.
 	ECHOffered bool
 
+	// PeerTLSFlags is the set of TLS Flags sent by the Peer.
+	PeerTLSFlags []TLSFlag
+
+	// AgreedTLSFlags is the set of TLS Flags mutually supported by the Client
+	// and Server.
+	AgreedTLSFlags []TLSFlag
+
+	// RequestClientCert is true if the server decided to request a client
+	// certificate.
+	RequestClientCert bool
+
 	// ekm is a closure exposed via ExportKeyingMaterial.
 	ekm func(label string, context []byte, length int) ([]byte, error)
 }
@@ -467,6 +479,20 @@ const (
 	// Legacy signature and hash algorithms for TLS 1.2.
 	PKCS1WithSHA1 SignatureScheme = 0x0201
 	ECDSAWithSHA1 SignatureScheme = 0x0203
+)
+
+// TLSFlag is the index of a bit in the TLSFlags bit array that should be set to
+// send the flag. Because TLSFlags can have so many different uses, many yet to
+// be defined, this extension should not be copied into the ECH
+// ClientHelloOuter.
+type TLSFlag uint16
+
+const (
+	// ExperimentalFlagSupportMTLS is a flag that signals that a client supports
+	// mTLS, and will be able to respond to CertificateRequest messages
+	// appropriately.
+	// https://datatracker.ietf.org/doc/draft-jhoyla-req-mtls-flag/
+	ExperimentalFlagSupportMTLS TLSFlag = 0x50
 )
 
 // ClientHelloInfo contains information from a ClientHello message in order to
@@ -905,6 +931,12 @@ type Config struct {
 	// See https://tools.ietf.org/html/draft-ietf-tls-subcerts.
 	SupportDelegatedCredential bool
 
+	// TLSFlagsSupported is the list of flags that the client or server is
+	// willing to support. This is currently limited to the set of flags set in
+	// the ClientHello, although the draft specifies various other messages
+	// where they can appear.
+	TLSFlagsSupported []TLSFlag
+
 	// mutex protects sessionTicketKeys and autoSessionTicketKeys.
 	mutex sync.RWMutex
 	// sessionTicketKeys contains zero or more ticket keys. If set, it means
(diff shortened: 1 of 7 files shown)


bwesterb:
Confirmed except for the new commit it's a rebase.

```
$ git range-diff cf/jhoyla/tls-flags^..cf/jhoyla/tls-flags  cf/cf..jonathan/jhoyla/tls-flags-rebase 
1:  4092d3f479 = 1:  f46052fa94 Implement TLS Flags and the request mTLS flag.
-:  ---------- > 2:  5324115d20 Fix nits
```

bwesterb (approved):

src/crypto/tls/common.go
```diff
@@ -364,6 +365,15 @@ type ConnectionState struct {
 	// This means the client has offered ECH or sent GREASE ECH.
 	ECHOffered bool
 
+	// PeerTLSFlags is the set of TLS Flags sent by the Peer.
+	PeerTLSFlags []TLSFlag
+
+	// AgreedTLSFlags is the set of TLS Flags mutually supported by the Client and Server
```
cjpatton: Also, wrap comments at 80 columns (here and below)
```suggestion
	// AgreedTLSFlags is the set of TLS Flags mutually supported by the Client and Server.
```

src/crypto/tls/common.go
```diff
@@ -364,6 +365,15 @@ type ConnectionState struct {
...
+	AgreedTLSFlags []TLSFlag
+
+	// RequestClientCert is true if the server decided to request a client certificate
```
cjpatton:
```suggestion
	// RequestClientCert is true if the server decided to request a client certificate.
```

src/crypto/tls/common.go
```diff
@@ -469,6 +479,20 @@ const (
 	ECDSAWithSHA1 SignatureScheme = 0x0203
 )
 
+// A TLSFlag is the index of the bit in the TLSFlags bit array that should be
```
cjpatton: Go comment conventions.
```suggestion
// TLSFlag is the index of a bit in the TLSFlags bit array that should be
```

src/crypto/tls/common.go
```diff
@@ -469,6 +479,20 @@ const (
...
+// set to send the flag. Because TLSFlags can have so many different uses, many
+// yet to be defined, this extension should not be copied into the ECH
+// ClientHelloOuter
```
cjpatton:
```suggestion
// ClientHelloOuter.
```

src/crypto/tls/common.go
```diff
@@ -469,6 +479,20 @@ const (
...
+type TLSFlag uint16
+
+const (
+	// ExperimentalFlagSupportMTLS is a flag that signals that a client
+	// supports mTLS, and will be able to respond to CertificateRequest
+	// messages appropriately.
```
cjpatton: Would it be more correct to say that the flag "prompts" the server to send a CertificateRequest?

src/crypto/tls/handshake_messages.go
```diff
@@ -577,6 +586,12 @@ func (m *clientHelloMsg) unmarshal(data []byte) bool {
 				}
 				m.supportedVersions = append(m.supportedVersions, vers)
 			}
+		case extensionTLSFlags:
+			var flagsList cryptobyte.String
+			if !extData.ReadUint8LengthPrefixed(&flagsList) || flagsList.Empty() {
```
cjpatton, line 591: Just wanted to confirm that it's illegal to send an empty flags list?

src/crypto/tls/handshake_server_test.go
```diff
@@ -539,6 +539,41 @@ func testCrossVersionResume(t *testing.T, version uint16) {
...
+			PrivateKey:  testRSAPrivateKey,
+		}},
+		TLSFlagsSupported: []TLSFlag{0x50},
+	}
+	clientCert, err := X509KeyPair([]byte(clientECDSACertificatePEM), []byte(clientECDSAKeyPEM))
+	if err != nil {
+		t.Fatalf("couldn't load client certs")
+	}
+	clientConfig := &Config{
+		TLSFlagsSupported:  []TLSFlag{0x50},
```
cjpatton: Suggestion: Generate a flags list with multiple flags in it. Also, it might be good to choose flags at random so that we excercise the encoding more comprehensively.

src/crypto/tls/handshake_server_test.go
```diff
@@ -539,6 +539,41 @@ func testCrossVersionResume(t *testing.T, version uint16) {
...
+	}
+	if state.AgreedTLSFlags[0] != TLSFlag(0x50) {
+		t.Fatalf("Failed to agree correct flags")
+	}
+	if !state.RequestClientCert {
+		t.Fatalf("Failed to request client cert")
+	}
+	if len(state.PeerCertificates) == 0 {
+		t.Fatalf("Didn't receive correct client certs")
+	}
```
cjpatton, line 624: Is `state` the client's state? If so, this test should also confirm that the server has received the flags sent by the client.

src/crypto/tls/handshake_server_tls13.go
```diff
@@ -892,6 +942,11 @@ func (hs *serverHandshakeStateTLS13) sendServerParameters() error {
 }
 
 func (hs *serverHandshakeStateTLS13) requestClientCert() bool {
+	for _, flag := range hs.tlsFlags {
+		if flag == ExperimentalFlagSupportMTLS {
+			return true
+		}
+	}
```
cjpatton, line 949: Suggestion, feel free to ignore: I feel like it ought to be possible to do this without iterating over the set flags. Can't we just do some bit fiddling on the the encoded TLS flags to see if the req mTLS flag is set?

cjpatton, line 949: I'd suggest adding a test that demonstrates request mTLS when triggered by this flag.

cjpatton (approved):
Looks good! I have some comments on style, plus some improvements for testing I'd strongly recommend. Feel free to merge when you're happy.

src/crypto/tls/common.go
jhoyla: Technically no, because the server MAY ignore it, even if it supports it. DavidBen persuaded me to keep it phrased like this so that it was always clear what to do. It does mean that req mTLS is possibly not a great name any more.

src/crypto/tls/handshake_messages.go
jhoyla: Yup, see https://datatracker.ietf.org/doc/html/draft-ietf-tls-tlsflags-15#section-2-11
> An implementation that receives an all-zero value for this extension or a value that contains trailing zero bytes MUST generate a fatal illegal_parameter alert.

src/crypto/tls/handshake_server_test.go
jhoyla: Weirdly the order of the returned parameters is `server_state, client_state, err`.

PR #192, opened 2026-04-19
crypto/tls: add ECH provider rejection unit tests

Abhicodeitout:
## Summary
Add unit tests for ECH provider rejection behavior when the client-facing server receives malformed GREASE-like or unsupported algorithm combinations.

## What changed
- Added a test case that uses a known config_id with an encapsulated key generated under a different KEM.
  - Expected behavior: provider returns reject (not abort).
- Added a test case that uses a known config_id with an unsupported KDF/AEAD pair.
  - Expected behavior: provider returns reject.
- Kept all changes in src/crypto/tls/ech_test.go.

## Why
A reviewer suggestion on #95 asked for explicit coverage that the client-facing server rejects:
- GREASE ECH that matches config_id but uses a different KEM.
- Unsupported KDF/AEAD algorithm combinations.

These tests lock in the intended reject path for both cases.

## Verification
Executed with repository toolchain:

GOROOT=/workspaces/go ../bin/go test crypto/tls -run TestECHProvider -count=1 -v

Result: PASS, including both new subtests.

## Commit
Includes Signed-off-by: Abhishek Kumar Kushwaha <abhithegabbar@gmail.com>.

### DIFF
--- a/src/crypto/tls/ech_test.go
+++ b/src/crypto/tls/ech_test.go
@@ -13,6 +13,8 @@ import (
 	"fmt"
 	"testing"
 	"time"
+
+	"github.com/cloudflare/circl/hpke"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


Abhicodeitout:
@benburkert  @evanphx @bmizerany  can you review

